PR #33, opened 2014-01-02
CRM-13936 - In Wordpress, CiviCRM blocks the execution of the Edit function in the Wordpress media library with error "call to undefined function ts()"

dlobo:
- Added an initialization function in the function add_form_button_html
- Made more of the code conform to civi coding standards (it was a mixture previously)
- eliminated lots of white space
...

http://issues.civicrm.org/jira/browse/CRM-13936


colemanw:
Seems like this should be closed since it is against 4.4 branch. If still needed it should be re-done against master.

PR #64, opened 2014-11-13
Keep WordPress functional during install process

christianwach:
Fixes https://issues.civicrm.org/jira/browse/CRM-14692

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -92,6 +92,13 @@ if (!defined('CIVICRM_SETTINGS_PATH')) {
   define( 'CIVICRM_SETTINGS_PATH', CIVICRM_PLUGIN_DIR . 'civicrm.settings.php' );
 }
 
+// test if Civi is installed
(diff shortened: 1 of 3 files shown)


christianwach:
And another report:
http://forum.civicrm.org/index.php/topic,35340.0.html

Do any committers have time to merge this, #65 and #66?

kcristiano:
@christianwach since this is affecting version back to 4.4.x (at least if I am reading the reports correctly) can we also apply this to the 4.4 and 4.5 branches?  I am just not sure if this PR is dependent on previous commits to 4.6

christianwach:
@kcristiano This PR won't merge with 4.5 or 4.4, but I'll create separate PRs for those shortly

christianwach:
@kcristiano Having said that, the this PR and the other two I mentioned should really be merged before 4.6 or there'll be inevitable issues raised because of them.

kcristiano:
@christianwach Agreed, I would like to see this in as early as we can

PR #72, opened 2015-02-09
Load `wp_head` scripts when on Civi admin page (CRM-15937)

christianwach:
Solves https://issues.civicrm.org/jira/browse/CRM-15937

@kcristiano @colemanw I can confirm that the JS and CSS fail to load once Civi is installed. This PR fixes the script loading. I'll take a look at the default for 4.5.x as well.

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -565,7 +565,7 @@ class CiviCRM_For_WordPress {
       );
 
       // add top level menu item
-      add_menu_page(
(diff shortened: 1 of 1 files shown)


christianwach:
I can confirm that all seems to be well with the 4.5 install process.

kcristiano:
@christianwach Thanks for this.  It works for me as well on the 4.4 branch.

PR #73, opened 2015-04-29
CRM-16408 WordPress Installer changes WIP

kcristiano:
First pass at WordPress installer changes.   Enable checking both existing and new location for civicrm.settings.php.  Also includes minor wp-cli change by PalanteJon.

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -89,8 +89,17 @@ if (!defined( 'CIVICRM_PLUGIN_DIR')) {
 
 // store PATH to this plugin's settings file
 if (!defined('CIVICRM_SETTINGS_PATH')) {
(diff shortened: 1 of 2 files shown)


christianwach:
@kcristiano It'd be nice to stick to either tab indenting *or* double-space indenting. I far prefer tab indented, but I think it'll irritate the hardcore Civi folks :disappointed:

Wasn't there also an objection to the timezone change at some point? I'll have a dig around.

kcristiano:
@christianwach Yes, I realized now that my indenting is not consistent and must be before merge.  There has been timezone discussion,  After @andrewhunt and I reviewed the WP behavior once the timezone_string option is set CiviCRM reports no errors.  This is even if the TZ is many hours different than PHP and MySQL.  Looks to be a bug, just not sure whose.

kcristiano:
Close in favor of https://github.com/civicrm/civicrm-wordpress/pull/85

PR #79, opened 2015-05-26
Fix basepage title (master)

christianwach:
PR in response to the issue raised on [StackExchange](http://civicrm.stackexchange.com/questions/2605/wordpress-page-title-only-fixed-in-4-6-4). This fixes things for me, even when Open Graph plugins try to hijack the title.


colemanw:
@christianwach commits to the 4.6 branch are still being forward-merged on a weekly basis so unless the code for master is different, there's no need for you to submit twice.
FYI our policy is to always forward-merge the latest stable branch into master. So 4.6 commits will continue being merged until the release of 4.7.

christianwach:
@colemanw I didn't know that - thanks for the info.

PR #81, opened 2015-06-10
CRM-16650 - Rewrite shortcode form

colemanw:
https://issues.civicrm.org/jira/browse/CRM-16650


christianwach:
@colemanw Does this PR have a corresponding civicrm-core PR? Otherwise, the shortcodes will stop working completely...

kcristiano:
@christianwach https://github.com/civicrm/civicrm-core/pull/5987  is the corresponding civicrm-core PR.

christianwach:
@kcristiano Thanks

colemanw:
Sorry about the noisy commit, my IDE flagged a bunch of formatting stuff and I couldn't help myself...

christianwach:
@colemanw Yeah, and your IDE is set to Civi standards, not WordPress ones :(

civicrm.php
```diff
@@ -507,6 +519,9 @@ public function register_hooks_admin() {
     // set page title
     add_filter( 'admin_title', array( $this, 'set_admin_title' ) );
 
+    // print CiviCRM's header
+    add_action('admin_head', array( $this, 'wp_head' ), 50);
```
colemanw: @christianwach this is the 1 *real* change to this file, and I wanted to run it by you. It took me a while to figure out why my call to `addCoreResources` wasn't working in `civicrm.shortcodes.modal.php` and finally realized this wasn't being called. So instead of doing it from there, I thought it would be more developer-friendly to have this action run unconditionally (note removal of conditional action below). If Civi isn't bootstrapped or hasn't added anything to the header, the function does nothing, so seems safe to me.

colemanw:
I think we're going to have to agree to disagree about coding standards ;)

christianwach:
> I think we're going to have to agree to disagree about coding standards ;)

I agree ;-) Although this is a *WordPress* plugin, isn't it?

/me gets coat

civicrm.php
christianwach: Yes, I saw that you'd relocated it. I can't give an opinion on it until I've had a chance to test. That may not be until next week, I'm afraid. Deadline looming *sigh*

colemanw:
;)
I've merged this so other development can move forward. Your testing and feedback would still be useful though. Thanks.

christianwach:
Hmm. It's slightly awkward that I can no longer use master on a stable version of Civi - makes it harder to work on this plugin. When is https://github.com/civicrm/civicrm-core/pull/5987 going into a Civi distributable?

PR #85, opened 2015-08-30
CRM-16408  

kcristiano:
Move civicrm.settings.php out of code directory. Place directory in proper location for WP repository standards.  Enable WP check to look in new location. Support Deprecated Settings file location as well as new location

CRM-16408: Installer rewrite https://issues.civicrm.org/jira/browse/CRM-16408

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -582,9 +582,10 @@ class CiviCRM_For_WordPress {
       // check for settings
       if ( ! CIVICRM_INSTALLED ) {
         $error = FALSE;
-      } else {
+      } elseif ( file_exists( CIVICRM_SETTINGS_PATH) ) {
         $error = include_once ( CIVICRM_SETTINGS_PATH );
-        $error =  $error . include_once ( CIVICRM_SETTINGS_PATH_DEPRECATED );
+      } elseif (file_exists( CIVICRM_SETTINGS_PATH_DEPRECATED  ) ) {
+        $error =  include_once ( CIVICRM_SETTINGS_PATH_DEPRECATED );
       }
 
       // autoload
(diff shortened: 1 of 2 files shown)


kcristiano:
Dependency : https://github.com/civicrm/civicrm-core/pull/6807

civicrm.php
```diff
@@ -89,14 +89,21 @@
...
 if (!defined('CIVICRM_SETTINGS_PATH')) {
-  define( 'CIVICRM_SETTINGS_PATH', CIVICRM_PLUGIN_DIR . 'civicrm.settings.php' );
+  $upload_dir    = wp_upload_dir();
+  $wp_civi_settings = $upload_dir['basedir'] . DIRECTORY_SEPARATOR . 'civicrm' . 'civicrm.settings.php' ;
+  define( 'CIVICRM_SETTINGS_PATH', $upload_dir['basedir'] . DIRECTORY_SEPARATOR . 'civicrm'. DIRECTORY_SEPARATOR . 'civicrm.settings.php' );
+}
+
+// store 4.6 and prior version PATH to this plugin's settings file
+if (!defined('CIVICRM_SETTINGS_PATH_DEPRECATED')) {
+  define( 'CIVICRM_SETTINGS_PATH_DEPRECATED', CIVICRM_PLUGIN_DIR . 'civicrm.settings.php' );
```
totten: I think there are parts of the core code which rely on `CIVICRM_SETTINGS_PATH` pointing at the real/live copy of `civicrm.settings.php`. (Ex: The upgrader sometimes needs to patch the file for a new version.) In this code, I think we get the wrong value for legacy sites. Would it be ok to first do the check to determine the path -- and then set the constant CIVICRM_SETTINGS_PATH?

kcristiano:
https://github.com/civicrm/civicrm-wordpress/pull/87  rebased to latest master.  see https://github.com/civicrm/civicrm-core/pull/6926

PR #96, opened 2016-02-29
Do not force "page.php" as basepage template

christianwach:
Jira ticket here: https://issues.civicrm.org/jira/browse/CRM-18113

The comments in the code explain the updated logic. tl;dr CiviCRM is being over-zealous in overriding the chosen template.


colemanw:
Has this been tested out on a few configurations? Happy to merge if so.
@kcristiano @danaskallman ?

kcristiano:
Yes we have done so and we have it in production.

PR #103, opened 2016-06-08
CRM-16421 CRM-17633 WIP Changes to support WP in it's own directory

kcristiano:
Syncronise get_base_url funcyion in civicrm.php with CRM_Utils_System_WordPress

---

 * [CRM-16421: Work to get CiviCRM for WordPress in WordPress' official Repository](https://issues.civicrm.org/jira/browse/CRM-16421)
 * [CRM-17633: WordPress in own directory breaks CiviCRM](https://issues.civicrm.org/jira/browse/CRM-17633)

PR #105, opened 2016-09-25
CRM-17633 WIP Changes to support WP in it's own directory  …

kcristiano:
… Replaced #103

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -1413,6 +1413,12 @@ class CiviCRM_For_WordPress {
    * Clone of CRM_Utils_System_WordPress::getBaseUrl() whose access is set to
    * private. Until it is public, we cannot access the URL of the basepage since
    * CRM_Utils_System_WordPress::url()
+   * 
+   * 27-09-2016
+   * CRM-16421 CRM-17633 WIP Changes to support WP in it's own directory
+   * https://wiki.civicrm.org/confluence/display/CRM/WordPress+installed+in+its+own+directory+issues
+   * For now leave hard coded wp-admin references.
+   * TODO: remove wp-admin references and replace with admin_url() in the future.  Look at best way to get path to admin_url
    *
    * @param bool $absolute Passing TRUE prepends the scheme and domain, FALSE doesn't
    * @param bool $frontend Passing FALSE returns the admin URL
--- a/civicrm.php
+++ b/civicrm.php
@@ -1420,29 +1420,26 @@ class CiviCRM_For_WordPress {
    * @return mixed|null|string
    */
   public function get_base_url($absolute, $frontend, $forceBackend) {
-    $config    = CRM_Core_Config::singleton();
-
-    if (!isset($config->useFrameworkRelativeBase)) {
-      $base = parse_url($config->userFrameworkBaseURL);
-      $config->useFrameworkRelativeBase = $base['path'];
+    $config = CRM_Core_Config::singleton();
+    if (!defined('CIVICRM_UF_ADMINURL')) {
+      define('CIVICRM_UF_ADMINURL', CIVICRM_UF_BASEURL . 'wp-admin/');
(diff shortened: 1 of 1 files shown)


kcristiano:
dependency- https://github.com/civicrm/civicrm-core/pull/9106

christianwach:
@kcristiano Thanks for working on this. I have a couple of queries:

* I see that the `CRM_Utils_System_WordPress::getBaseUrl()` and `CiviCRM_For_WordPress::get_base_url()` methods are now identical. Would it not be simpler to change the method access for `CRM_Utils_System_WordPress::getBaseUrl()` to `public` and then we could dispense with `get_base_url()` entirely?
* Is there a reason for using [a hard-coded directory name](https://github.com/civicrm/civicrm-wordpress/pull/105/commits/8c89681fcbb058c354bad8f2ca2bd5098dcc454b#diff-d1bc6ad2b4d05fe1a841fce94b11b8dbR1425)? There are a number of ways of changing from 'wp-admin' to something else: e.g. [this plugin](https://wordpress.org/plugins/protect-wp-admin/) and [this technique](https://wordpress.org/support/topic/how-to-change-the-admin-url-or-wp-admin-to-secure-login/).
* Finally, is there any assumption about whether or not WordPress is bootstrapped when either `CRM_Utils_System_WordPress::getBaseUrl()` or `CiviCRM_For_WordPress::get_base_url()` is called? If WordPress is *always* bootstrapped, then wouldn't it make more sense to use native WordPress functions to discover the URL?

Cheers, Christian

kcristiano:
@christianwach Thanks for the feedback.

- it's possible we could change the method - I am trying to keep compatibility with other code.  Backwards compatibility and all :)

- I am using 'wp-admin' hardcoded as my attempts to get it to return via the admin_url failed due to issues with the bootstrap process.   

I worked yesterday on all edge cases - changing uploads folder, changing plugins folder and did not even attempt to change wp-admin.  I think we need to draw a line and decide what we can support.   The possible permutations are causing issues - mostly from 'extern' calls and when the CMS is not boostrapped.

- Lastly we are assuming WP is not bootstrapped - hence why I want the values defined in the civicrm.settings.php file.   Tim's civicrm.config.php can find the settings file and we need those paths for the bootsrap process.   

Perhaps I am getting 'testing blind' over this - so if you see a way to improve that would be great.

christianwach:
> we are assuming WP is not bootstrapped

@kcristiano thanks for clarifying. If that's the case, then my only request would be for more documentation - I'm unclear where the various constants are set (or could be set) so it would be really helpful to leave some breadcrumbs for developers in the respective method docblocks.

Cheers, Christian

kcristiano:
@christianwach added a docblock - I also updated https://wiki.civicrm.org/confluence/display/CRM/WordPress+installed+in+its+own+directory+issues  with all the testing I have done.

The next step is fixing the installer to handle this and more testing.   We will need to update README.md and readme.txt to cover these changes.

alifrumin:
I tested this and it worked for me!

kcristiano:
jenkins test this please

kcristiano:
https://github.com/civicrm/civicrm-core/pull/10214  is dependent on this PR

kcristiano:
@totten tested and merged your improvements.   Works fine.  Would like this in early in an RC cycle to get more eyes/hands on the rela world testing.

kcristiano:
https://github.com/civicrm/civicrm-core/pull/11031  is now the matching PR in civicrm-core

alifrumin:
I tested this pr and https://github.com/civicrm/civicrm-core/pull/11031 on a wordpress install with wordpress in the root site directory and with wordpress in its own directory and did not get any issues regarding directories or resource urls.

eileenmcnaughton:
Merging per related PR #11031

PR #106, opened 2016-10-20
CRM-18062 backport

agileware:
Just the same commits from 4.7 - applied and tested.

---
- [CRM-18602](https://issues.civicrm.org/jira/browse/CRM-18602)

---
- [CRM-18062: Mailing processing ignores timezone, and sends mailings at wrong time](https://issues.civicrm.org/jira/browse/CRM-18062)


kcristiano:
Thanks @agileware I can confirm that this patch is working on 4.6.22

yashodha:
Merging as per confirmation of the backport in the comment

PR #107, opened 2016-11-18
CRM-19665 WordPress: set canonical URL when using basepage

agh1:
* [CRM-19665: Canonical URL for WP basepage pages is the basepage itself](https://issues.civicrm.org/jira/browse/CRM-19665)

### DIFF
--- a/includes/civicrm.basepage.php
+++ b/includes/civicrm.basepage.php
@@ -76,6 +76,12 @@ class CiviCRM_For_WordPress_Basepage {
       return;
     }
 
+    // In WP 4.6.0+, tell it URL params are part of canonical URL
+    add_filter( 'get_canonical_url', array($this, 'basepage_canonical_url' ), 999);
+
+    // Yoast SEO has separate way of establishing canonical URL
+    add_filter( 'wpseo_canonical', array($this, 'basepage_canonical_url' ), 999);
+
     // regardless of URL, load page template
     add_filter( 'template_include', array( $this, 'basepage_template' ), 999 );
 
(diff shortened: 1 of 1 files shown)


kcristiano:
commented on JIRA issue https://issues.civicrm.org/jira/browse/CRM-19665

christianwach:
Likewise commented on Jira. My only observations on the PR itself are minor and somewhat pedantic, but these things tend to build up over time. I'd prefer less idiosyncratic (or perhaps more consistent) use of whitespace, though I recognise that the code styling of this plugin favours CiviCRM's codestyle over that of WordPress. I also note that there's a missing `@param` description in the new method's docblock.

kcristiano:
Good point about the docblock.

I have tested this on WP 4.5.4 and WP 4.6.1 and it solves the problem described.

I think we should add All in One SEO support alongside Yoast SEO support.   Commented with the code we would need to add.

agh1:
Those recent changes can be ignored--@kcristiano made a PR against the branch that I didn't see till today, and I merged it not remembering that the change was already included.

kcristiano:
I commented here: https://issues.civicrm.org/jira/browse/CRM-19665

Tested again against current master branches, works propoerly.  Tested with Yoast SEO, AIOSEO and with no SEO plugin.   Works in all conditions.

I consider this a bug fix, and would like to see this patch included with 4.7.17

kcristiano:
@colemanw @totten This is ready to merge.  Can you review?

kcristiano:
@eileenmcnaughton since you have been watching this repo lately, can you take a look at this?   It's ready to merge, having this in the next release (4.7.18) would be great.

Thanks

PR #111, opened 2017-03-07
CRM-20225 Add classes to body element when on basepage

christianwach:
As described on [Jira](https://issues.civicrm.org/jira/browse/CRM-20225) and [Stack Exchange](http://civicrm.stackexchange.com/questions/17403/html-event-pages-unspecific-ids-and-classes-at-the-top/), this implements the addition of generic classes to the body tag when on the WordPress basepage.

### DIFF
--- a/includes/civicrm.basepage.php
+++ b/includes/civicrm.basepage.php
@@ -178,6 +178,9 @@ class CiviCRM_For_WordPress_Basepage {
     // tweak admin bar
     add_action( 'wp_before_admin_bar_render', array( $this->civi, 'clear_edit_post_menu_item' ) );
 
+    // add body classes for easier styling
+    add_filter( 'body_class', array( $this, 'add_body_classes' ) );
+
     // flag that we have parsed the base page
     $this->basepage_parsed = TRUE;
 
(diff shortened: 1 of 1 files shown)


christianwach:
Should be possible to apply this PR to previous versions of CiviCRM without modification too.

eileenmcnaughton:
@kcristiano any chance you can review this?

kcristiano:
@christianwach I'll do some testing on here, it would help if you had some examples of what the classes should look like after applying the patch.  On a quick before and after comparison the output looks identical.

christianwach:
@kcristiano I guess there is the assumption that a theme implements the `body_class()` function, but I believe that it's a requirement for the WordPress Theme Directory.

For me with any of the "Twenty*" themes, I see (for example) `civicrm civicrm-event civicrm-event-ical` added to the `<body>` tag when viewing 'http://domain.whatever/civicrm/?page=CiviCRM&q=civicrm/event/ical&reset=1&list=1&html=1'. If I visit '/civicrm/?page=CiviCRM&q=civicrm/pcp/info&reset=1&id=1', I see `civicrm civicrm-pcp civicrm-pcp-info` added.

Essentially what should be happening is that the `q=` arguments are parsed and the sequence of increasing-specific classes are added, meaning that CSS selectors can be added at whatever level of granularity is required.

kcristiano:
@christianwach I need more coffee before I test.  Thank you, I was looking at the div classes not the body tag.

This works nicely and I agree it should be merged.

@eileenmcnaughton OK to merge

christianwach:
@eileenmcnaughton Thanks! I'll report back to the chap on SE.

PR #112, opened 2017-05-14
Enable CiviCRM Extensions to interact with the WordPress shortcode

christianwach:
Fixes [CRM-18754](https://issues.civicrm.org/jira/browse/CRM-18754). WordPress plugins and CiviCRM Extensions can now process custom parameters in the `civicrm` shortcode.

Let's take the example shortcode given in the Jira ticket:

`[civicrm component="contribution" id="2" mode="live" discount="pkpwpabs7" hijack="0"]`

Here's some code that an extension can use to receive data from the custom shortcode attribute and act on it:

```php
if ( function_exists( 'add_filter' ) ) {
	add_filter( 'civicrm_shortcode_preprocess_atts', 'extensionprefix_amend_args', 10, 2 );
	add_filter( 'civicrm_shortcode_get_data', 'extensionprefix_amend_data', 10, 3 );
}

/**
 * Filter the CiviCRM shortcode arguments.
 *
 * Modify the attributes that the 'civicrm' shortcode allows. The attributes
 * that are injected (and their values) will become available in the $_REQUEST
 * and $_GET arrays.
 *
 * @param array $args Existing shortcode arguments
 * @param array $shortcode_atts Shortcode attributes
 * @return array $args Modified shortcode arguments
 */
function extensionprefix_amend_args( $args, $shortcode_atts ) {

	// our custom attribute name & default
	$name = 'discount';
	$default = 'foo';

	// add either passed value or default
	if ( array_key_exists($name, $shortcode_atts) ) {
		$args[$name] = $shortcode_atts[$name];
	} else {
		$args[$name] = $default;
	}

	return $args;

}

/**
 * Filter the CiviCRM shortcode data array.
 *
 * Let's add some arbitrary text to the pre-rendered shortcode's description to
 * indicate that this extension has something to say.
 *
 * @param array $data Existing shortcode data
 * @param array $atts Shortcode attributes array
 * @param array $args Shortcode arguments array
 * @return array $data Modified shortcode data
 */
function extensionprefix_amend_data( $data, $atts, $args ) {

	// our custom attribute name
	$name = 'discount';

	// add some arbitrary text to pre-rendered shortcode
	if ( array_key_exists($name, $atts) ) {
		$data['text'] .= ' ' . ts('Discount code: ') . $atts[$name];
	}

	return $data;

}
```

In addition, more sophisticated plugins and extensions can also filter the parameters passed the the CiviCRM API when there are multiple shortcodes present. This would allow them to retrieve additional (or alternative) data for display in the pre-rendered shortcode.

### DIFF
--- a/includes/civicrm.shortcodes.php
+++ b/includes/civicrm.shortcodes.php
@@ -145,6 +145,11 @@ class CiviCRM_For_WordPress_Shortcodes {
     // did we get any?
     if ( $shortcodes_present ) {
 
+      // we need CiviCRM initialised prior to parsing shortcodes
+      if (!$this->civi->initialize()) {
+        return;
+      }
+
       // how should we handle multiple shortcodes?
       if ( $shortcodes_present > 1 ) {
 
@@ -555,11 +560,11 @@ class CiviCRM_For_WordPress_Shortcodes {
    * Preprocess CiviCRM-defined shortcodes
    *
    * @param array $atts Shortcode attributes array
-   * @return void
+   * @return array $args Shortcode arguments array
    */
   public function preprocess_atts( $atts ) {
 
-    extract( shortcode_atts( array(
+    $shortcode_atts = shortcode_atts( array(
       'component' => 'contribution',
       'action' => NULL,
       'mode' => NULL,
@@ -569,8 +574,11 @@ class CiviCRM_For_WordPress_Shortcodes {
       'cs' => NULL,
       'force' => NULL,
       ),
-      $atts
-    ) );
+      $atts,
+      'civicrm'
+    );
+
+    extract( $shortcode_atts );
 
     $args = array(
       'reset' => 1,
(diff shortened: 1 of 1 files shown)


kcristiano:
@christianwach This looks good.   Tested 4.7.26/master and WP 4.8.2 

Applied cleanly and was able to test with existing shortcodes and with your code example.

Existing shortcodes work as expected.   Adding params work properly as well.

Ready to commit

kcristiano:
@colemanw if you can review it'd be appreciated.

kcristiano:
@totten @colemanw any chance you could take a look, we'd like to see this merged, it's be here for a while and tested.   Let me know if anything else is needed to speed merging along

JoeMurray:
@monishdeb could you do a technical review. Personally I would like to see this functionality in core, but I have not reviewed PR.

colemanw:
That would be great. Thanks @JoeMurray & @monishdeb

includes/civicrm.shortcodes.php
```diff
@@ -569,8 +574,11 @@ public function preprocess_atts( $atts ) {
       'cs' => NULL,
       'force' => NULL,
       ),
-      $atts
-    ) );
+      $atts,
+      'civicrm'
```
monishdeb: Legit, signifies ```civicrm``` shortcode and allows it add filter args.

monishdeb:
Great work @christianwach and thanks for your feedback @kcristiano 

I tested with a dummy extension to influence civicrm shorcodes using the new filters. From code pov it's a straightforwad and necessarychange which doesn't cause any unintended issue. Merging it

We should document this improvement but not sure where at https://github.com/civicrm/civicrm-dev-docs/tree/master/docs .

ping @seanmadsen @JoeMurray

christianwach:
@monishdeb Thanks!

seanmadsen:
Thanks for thinking of docs, @monishdeb. I added https://github.com/civicrm/civicrm-dev-docs/pull/428 as a starting point for the docs by copy-pasting the description in the beginning of this PR. Can someone help finish those docs? I don't understand enough about the PR to feel confident that my copy-paste is actually good documentation. 

Side note: in general I'd really like to see us ensure that any necessary documentation changes are completed *before* merging changes. Otherwise things tend to fall through the cracks.

PR #113, opened 2017-08-20
CRM-21084: Extensibility improvement

GinkgoFJG:
Makes available to WP devs information about the shortcodes in use on the current page.

---

 * [CRM-21084: Give themers and plugin devs ability to test if CiviCRM is currently being displayed in a WordPress shortcode](https://issues.civicrm.org/jira/browse/CRM-21084)

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -134,6 +134,20 @@ class CiviCRM_For_WordPress {
    */
 
   /**
+   * Shortcodes active in the current page.
+   *
+   * @var array
+   */
+  private static $civicrm_active_shortcodes = array();
+
+  /**
+   * Flag used to indicate whether or not CiviCRM's shortcodes have been parsed yet.
+   *
+   * @var bool
+   */
+  private static $civicrm_shortcodes_parsed = FALSE;
+
+  /**
    * @var CiviCRM_For_WordPress
    */
   private static $instance;
@@ -405,6 +419,25 @@ class CiviCRM_For_WordPress {
 
   }
 
+  /**
+   * Setter for shortcodes that are active in the page.
+   *
+   * Callback to WordPress 'civicrm_shortcodes_parsed' hook.
+   *
+   * @param array $shortcodes
+   */
+  public function civicrm_active_shortcodes_set($shortcodes) {
+    self::$civicrm_active_shortcodes = $shortcodes;
+    self::$civicrm_shortcodes_parsed = TRUE;
+  }
+
+  public function civicrm_active_shortcodes_get() {
+    return apply_filters('civicrm_active_shortcodes', self::$civicrm_active_shortcodes);
+  }
+
+  public function civicrm_shortcodes_parsed_get() {
+    return self::$civicrm_shortcodes_parsed;
+  }
 
   // ---------------------------------------------------------------------------
   // Files
@@ -516,6 +549,8 @@ class CiviCRM_For_WordPress {
    * @return void
    */
   public function register_hooks_common() {
+    // expose to extension the list of shortcodess in use on the current page
+    add_action('civicrm_shortcodes_parsed', array($this, 'civicrm_active_shortcodes_set'));
 
     // use translation files
     add_action( 'plugins_loaded', array( $this, 'enable_translation' ) );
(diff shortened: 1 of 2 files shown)


GinkgoFJG:
Hey, @christianwach. Thanks for looking at this. It occurred to me as I was drifting off to sleep one night after making this PR that it is not strictly necessary. Another way to approach the use case I described in CRM-21084 could go something like this:

```php
// listen for the civicrm_shortcodes_parsed event, then determine
// what kind of page we're on and manage filters accordingly
add_action('civicrm_shortcodes_parsed', 'manageCiviFilters');

function manageCiviFilters () {
  $isShortcode = (bool) count(civi_wp()->shortcodes->shortcodes);
  if ($isShortcode) {
    add_filter('the_content', 'fancyShortcodeFilter');
  }

  if ($isShortcode || civicrm_wp_in_civicrm()) {
    remove_filter('the_content', 'civiUnawareFilter');
  }
}

function fancyShortcodeFilter($content) {
  // remove some text we don't want to display in shortcode mode
  return $content;
}
```

This code works without any changes to core. In fact, it's preferable because you have guaranteed access to the shortcodes after the `civicrm_shortcodes_parsed` event, even if the listener has a very low priority. (In my PR I threw in the ugly exception to deal with someone trying to call the function before the shortcodes were available.)

I haven't written a lot of WordPress code; this PR is the result of about three hours of thrashing about, trying to get my bearings. That said, I'm pretty sure these events (or I guess in WordPress you call them "actions?") aren't documented. Perhaps rather than another convenience function, what this code needs is some developer documentation. Does that sound right to you?

christianwach:
@GinkgoFJG 

> It occurred to me as I was drifting off to sleep one night after making this PR that it is not strictly necessary

That's when I do my best work too :-)

You're almost there with your code. The problem is that `civicrm_shortcodes_parsed` fires during the execution of the `wp` action which is well before The Loop runs. The reason it does so is that the shortcodes have to be pre-rendered for the variety of situations in which they can be present - especially significant for archive pages (and when two shortcodes in the same post) because CiviCRM cannot run twice as a result of one HTTP request. This is why you'll see "preview blocks" in such situations.

I suspect, however, that the `civicrm_shortcodes_parsed` action is a bit of a red herring in your case. The function `has_shortcode()` makes your life much easier...

The thing with `wpautop` etc is that once you've unhooked these functions the content of the next item in The Loop will not have it applied. This may or may not be desired, but is easily accommodated so that the code works regardless of context.

**Warning untested code ahead** Try the following sort of thing out for size - note how the filters are optionally removed and optionally re-added by injecting callbacks before and after the default priority:

```php
// hook into 'the_content' before wpautop is applied at priority 10
add_action( 'the_content', 'myprefix_maybe_remove_content_filters', 9, 1 );

/**
 * Decide whether to allow the application of wpautop etc or not.
 *
 * @param str $content The existing content.
 * @return str $content The modified content.
 */
function myprefix_maybe_remove_content_filters( $content ) {

	// if the CiviCRM shorcode is present
	if ( has_shortcode( $content, 'civicrm' ) ) {

		// yes: remove wpautop
		remove_filter( 'the_content', 'wpautop' );

		// remove any other filters here...

		// we know this post doesn't want these filters, but the next one might
		// so now trigger re-adding them for the next post in The Loop
		add_action( 'the_content', 'myprefix_readd_content_filters' ), 11, 1 );

	}
	
	// always return content
	return $content;
	
}

/**
 * Re-apply wpautop and other filters.
 *
 * @param str $content The existing content.
 * @return str $content The modified content (though not in this case).
 */
function myprefix_readd_content_filters( $content ) {

	// add wpautop
	add_filter( 'the_content', 'wpautop' );

	// add any others back in here...

	// now remove this callback otherwise wpautop etc will be applied multiple times!
	remove_action( 'the_content', 'myprefix_readd_content_filters', 11 );

	// always return content
	return $content;
	
}
```

One last thing to note: whether this works as expected or not depends on the version of WordPress that you're using. Prior to WordPress 4.7, you had to do some little tricks to get things to behave properly. With 4.7+ everything works as expected. I won't bore you with the details, but if you're *really* keen, you could check out the (staggeringly complex) ticket that ultimately led to a solution for this:
https://core.trac.wordpress.org/ticket/17817

Cheers, Christian

kcristiano:
@christianwach @GinkgoFJG just looping back on this.  Do we want to consider this PR to be merged?   Reading through it looks like we can close this and create a Documentation task.   Am I correct?

christianwach:
@kcristiano I'd agree - it's more of a documentation issue. If you're using BuddyPress, there are couple of functions that can help with filter management: [`bp_remove_all_filters`](https://github.com/buddypress/BuddyPress/blob/master/src/bp-core/bp-core-theme-compatibility.php#L794-L850) and [`bp_restore_all_filters`](https://github.com/buddypress/BuddyPress/blob/master/src/bp-core/bp-core-theme-compatibility.php#L852-L905). These can be used in the code above where I say "remove any other filters here" and "add any others back in here".

christianwach:
@kcristiano I think this ought to be closed and a Documentation task opened instead. the code I provided should suffice.

PR #110, opened 2017-02-21
CRM-20148 Fix WP-CLI installer

agh1:
This pull request looks insane because of the style changes that I ran first because my editor was blowing up with PHP_CodeSniffer errors: `wp-cli/civicrm.php` was in neither CiviCRM/Drupal style nor WordPress style.  This puts it in WordPress style.

The real business in here comes in 5612d6fc9c30f4eb70508b3f32d209eca2b5ff85, with a couple additional things in 72202de3ba69f2c1ddb24ee8b5dd59e9fac277ff, so it would probably be easier to browse through those.

The result works with the following two commands:

- `wp civicrm install --dbhost=localhost --dbname=XXXX --dbpass=XXXX --dbuser=XXXX --site_url=example.com` when CiviCRM is extracted in place and already activated in WordPress, and
- `wp civicrm install --dbhost=localhost --dbname=XXXX --dbpass=XXXX --dbuser=XXXX --site_url=example.com --zipfile=/path/to/civicrm-4.7.16-wordpress.zip --require=/somewhere/else/wp-cli/civicrm.php` when CiviCRM is not present but the WP-CLI file is at `/somewhere/else/wp-cli/civicrm.php`.

However, I have only run this on plain vanilla WordPress installs.

It would be nice to have @kcristiano's feedback to make sure it's consistent with setups I'm not aware of and with work he's doing to improve the installation process in WordPress.

---

 * [CRM-20148: WP-CLI installer won't run if it can be run](https://issues.civicrm.org/jira/browse/CRM-20148)

PR #114, opened 2017-08-31
Fix parameter parsing in wp-cli civicrm api.

jonesinator:
The entity and action were also being parsed as an invalid parameter.
This was fixed by shifting the args array after parsing them.

An unassigned $format variable was being referenced inside of the
default case for the switch statement if the --in option was neither
args nor json. This was fixed by assigning $format to the output of
getOption rather than switching on getOption directly.

The parameter matching regex had spaces added to it in pull request #110
that caused arguments of the format foo=bar to not be matched correctly,
so no parameters were actually being passed to the underlying api
method. The spaces were removed from the regex and the parsing worked.

Two other regex matches had spaces added to them by pull request #110
that I have not touched, but they seem suspect.

### DIFF
--- a/wp-cli/civicrm.php
+++ b/wp-cli/civicrm.php
@@ -174,7 +174,6 @@ if ( ! defined( 'CIVICRM_WPCLI_LOADED' ) ) {
 		 * Implementation of command 'api'
 		 */
 		private function api() {
-
 			$defaults = array( 'version' => 3 );
 
 			array_shift( $this->args );
@@ -182,8 +181,7 @@ if ( ! defined( 'CIVICRM_WPCLI_LOADED' ) ) {
 			array_shift( $this->args );
 
 			# parse $params
-
-			$format = $this->getOption( 'in', 'args' )
+			$format = $this->getOption( 'in', 'args' );
 			switch ( $format ) {
 
 				# input params supplied via args ..
--- a/wp-cli/civicrm.php
+++ b/wp-cli/civicrm.php
@@ -179,16 +179,18 @@ if ( ! defined( 'CIVICRM_WPCLI_LOADED' ) ) {
 
 			array_shift( $this->args );
 			list( $entity, $action ) = explode( '.', $this->args[0] );
+			array_shift( $this->args );
 
 			# parse $params
 
-			switch ( $this->getOption( 'in', 'args' ) ) {
+			$format = $this->getOption( 'in', 'args' )
+			switch ( $format ) {
 
 				# input params supplied via args ..
 				case 'args':
 					$params = $defaults;
 					foreach ( $this->args as $arg ) {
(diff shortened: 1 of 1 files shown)


totten:
Looks like this is your first patch for `civicrm-wordpress`. Thank you, @jonesinator!

A couple things:
 1. If @kcristiano or @christianwach approves, then we should merge. (Other reviewers welcome, but those guys are pretty sharp about the `civicrm-wordpress` patches.)
 2. Most PR's should reference a [JIRA](http://issues.civicrm.org/) issue (eg `CRM-XXXXX`). You can find some background about [issue tracking in the dev docs](https://docs.civicrm.org/dev/en/latest/tools/issue-tracking/) and then signup for an account at https://civicrm.org/user/register . (Warning: due to anti-spam measures, it can take a day or so for a new account to get setup. Feel free to ping if it's taking a while.)

kcristiano:
@jonesinator Thanks for the patch.

Can you detail the issue you are having with steps to reproduce?

I agree the spacing added in https://github.com/civicrm/civicrm-wordpress/pull/110 needs to be fixed.

I've applied the patch to my local dev env and I am getting the following error on any wp cv command:

```
PHP Parse error:  syntax error, unexpected 'switch' (T_SWITCH) in /srv/www/wpmaster/wp-content/plugins/civicrm/wp-cli/civicrm.php on line 187
Parse error: syntax error, unexpected 'switch' (T_SWITCH) in /srv/www/wpmaster/wp-content/plugins/civicrm/wp-cli/civicrm.php on line 187
```

I am testing against the latest master branch, I get this error wth both php 7.0 and php 5.6.  Any details on your environment would be helpful as well.

jonesinator:
Sorry, not sure how a semicolon disappeared between my testing and my commit. Your parsing error should be fixed by my latest patch. Also, apologies for not creating an issue in Jira. I'll try to follow the correct procedure with any changes in the future.

I'm only testing on php 7.0 at the moment. My issue is that any `wp-cli civicrm api` command that I pass arguments to in the format `key1=value1 key2=value2` using `--in=args` behaves as if no arguments are passed to it. The ability to pass `--in=json` works correclty. As I was debugging this issue I found the additional issues of use of the unassigned variable `$format` in the `default` case and the fact that the arguments were not being shifted prior to parsing, thus including the API entity and action being invoked as parameters.

For example, prior to the fix, the command `$ wp cv api contribution.get id=18` actually returned all contributions because the argument "id=10" is being ignored accidentally. After the fix it correctly only returns the contribution with an ID of 18.

kcristiano:
@jonesinator Thanks again for the patch and the details on how to reproduce.

Reproduced with the current civicrm.php file.   Applied patch.   Patch applies cleanly and now accepts arguments properly.

@totten This looks good to merge.  I'd target 4.7.25

totten:
Thank you, @jonesinator @kcristiano . Merging into `master` which will become 4.7.25.

PR #115, opened 2017-09-10
CRM-21166: WP-CLI commands fail due to whitespace in regex  Remove wh…

kcristiano:
…itespace from regex on parseDSN function.

### DIFF
--- a/wp-cli/civicrm.php
+++ b/wp-cli/civicrm.php
@@ -1136,7 +1136,7 @@ if ( ! defined( 'CIVICRM_WPCLI_LOADED' ) ) {
 
 			// Get phptype and dbsyntax
 			// $str => phptype( dbsyntax )
-			if ( preg_match( '|^( .+? )\( ( .*? )\ )$|', $str, $arr ) ) {
+			if ( preg_match( '|^(.+?)\((.*?)\)$|', $str, $arr ) ) {
 				$parsed['phptype']  = $arr[1];
 				$parsed['dbsyntax'] = ! $arr[2] ? $arr[1] : $arr[2];
 			} else {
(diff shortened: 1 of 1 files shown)


kcristiano:
This PR completes the cleanup in https://github.com/civicrm/civicrm-wordpress/pull/114  @totten is it possible to get this in 4.7.25 rc?

totten:
Yes, I think fixing a recent regression in the RC makes sense. Edited the PR to target the `4.7.25-rc`. Confirmed that this patch removes some `preg_match` errors when running `wp cv sql-dump`. If there's no other feedback in the next day, then let's merge into the RC.

(Note to self: after merging, we'll need to merge-forward to `master`.)

CC @jonesinator @agh1 because the JIRA description references related code from #110 and #114.

agh1:
Ugh sorry--thanks for catching this!  I suspect that it got introduced by the editor and/or PHPCS as it tries to enforce WP style with spaces inside of parentheses.

wp-cli/civicrm.php
```diff
@@ -1163,7 +1163,7 @@ private static function parseDSN( $dsn ) {
 
 			// Find protocol and hostspec
 
-			if ( preg_match( '|^( [^( ]+ )\( ( .*? )\ )/?( .*? )$|', $dsn, $match ) ) {
+			if ( preg_match( '|^([^(]+)\((.*? )\)/?(.*?)$|', $dsn, $match ) ) {
```
agh1: Is there one extra space to remove (after the first question mark)?

kcristiano: missed that.  Thanks.  Updating the PR

eileenmcnaughton:
@agh1 @totten looks like this has gone cold - good to merge?

agh1:
@eileenmcnaughton I think it is--I've applied these changes already in one spot where I encountered the bug.

PR #116, opened 2017-09-10
CRM-21168: Update CiviCRM documentation URLs.

mlutfy:
* [CRM-21168: WordPress plugin wiki\/support links are incorrect](https://issues.civicrm.org/jira/browse/CRM-21168)


kcristiano:
Patch looks good.  Applied, tested on install and existing site.   Works fine.  Ready to merge.

PR #117, opened 2017-09-26
4.7.25 rc

fwilly:


kcristiano:
@colemanw this was merged in 4.7.25-rc  and should also be merged to master.   Not certain we have a script for that.  If not this PR does need to be merged before 4.7.26-rc is cut

PR #118, opened 2017-09-30
CRM-21243 WordPress: logo in admin menu now uses SVG and works like others

agh1:
* [CRM-21243: Logo in WP menu doesn't follow standard](https://issues.civicrm.org/jira/browse/CRM-21243)


colemanw:
@kcristiano can you do a quick test of this?

kcristiano:
@colemanw @agh1 I like the look as we now fllow the standard.  Installed patch and works as expected.   For reference https://developer.wordpress.org/reference/functions/add_menu_page/ details that adding the svg  as a base64-encoded SVG using a data URI is compliant with the WP standards.

OK to merge from my perspective.

civicrm.php
```diff
@@ -737,17 +737,11 @@ public function enable_translation() {
...
+
```
totten: 👍 for SVG.

A slight nitpick... this does add 18kb of code in two source files, and they're duplicates. Would make sense to do something like:

```php
$civilog = file_get_contents(__DIR__ . "civilog.svg.b64");
```

or maybe

```php
$civilog = 'data:image/svg+xml;base64,' . base64_encode(file_get_contents(__DIR__ . "civilog.svg"));
```

agh1: I was thinking about the second, but I didn't want to require the server to encode the file every time the admin menu appears.  I never really thought about splitting the difference by going about it the first way, but I think I'll do that.

agh1:
@totten I just pulled the SVG to its own file and it appears to be working as well as before.

kcristiano:
@agh1 do we need file_get_contents since we are referencing an image file?

agh1:
@kcristiano yeah b/c it still needs to be an inline base64 image when it hits the browser.  the last commit is just a housekeeping measure for internal purposes.

colemanw:
Cool.

PR #120, opened 2017-12-02
CRM-21565: Change mkdir to use correct and more secure mode numbers

TBSliver:
The original setting of `777` is not a valid mode number for mkdir, and
should have been `0777` as it is an octal number. This has been changed
to `0755` to be in keeping with the wordpress permission scheme, which
is readable by all but only writeable by the current user. As these are
for backups a more strict setting may be preferable, but this is a
reasonable compromise

---

 * [CRM-21565: Change mkdir to use correct and more secure mode numbers](https://issues.civicrm.org/jira/browse/CRM-21565)

### DIFF
--- a/wp-cli/civicrm.php
+++ b/wp-cli/civicrm.php
@@ -654,7 +654,7 @@ if ( ! defined( 'CIVICRM_WPCLI_LOADED' ) ) {
 
 			$restore_backup_dir .= '/plugins/restore/' . $date;
 
-			if ( ! mkdir( $restore_backup_dir, 777, true ) ) {
+			if ( ! mkdir( $restore_backup_dir, 0755, true ) ) {
 				return WP_CLI::error( 'Failed creating directory: ' . $restore_backup_dir );
 			}
 
@@ -999,7 +999,7 @@ if ( ! defined( 'CIVICRM_WPCLI_LOADED' ) ) {
 			# begin upgrade
 
 			$backup_dir .= '/plugins/' . $date;
-			if ( ! mkdir( $backup_dir, 777, true ) ) {
+			if ( ! mkdir( $backup_dir, 0755, true ) ) {
 				return WP_CLI::error( 'Failed creating directory: ' . $backup_dir );
 			}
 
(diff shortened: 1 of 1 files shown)


kcristiano:
@TBSliver this looks good.  I've tested and it works on my system.  Before mergeing we should add a change on https://github.com/civicrm/civicrm-wordpress/blob/master/wp-cli/civicrm.php#L382  to set the permissions there to 0755 as well.   Can you make this change on your currebt branch so that this gets added to this PR?

I would like more testing on this due to the multiple hosting environments we encounter.   I am not sure why we set to 777 in the first place, so I do want to have this changed.   

Thanks for the work on this.

TBSliver:
@kcristiano done and done: https://issues.civicrm.org/jira/browse/CRM-21565

kcristiano:
@colemanw I am OK with merging this, but is there anyway to update the PR to include the JIRA issue for the release notes?  cc @agh1

TBSliver:
@kcristiano is this something that just needs editing on the original PR text, or is it something that gets put in on the merge notes when you accept and close the PR?

kcristiano:
@TBSliver I know that if the PR title includes the CRM-xxxx it helps the release notes get done.  If you can edit and add that in it would be great.  My hope is to get this in 4.7.30-RC as soon as possible

agh1:
Thanks @kcristiano @TBSliver: if you could please edit the title of this pull request to read

>  CRM-21565: Change mkdir to use correct and more secure mode numbers

that would be perfect!

TBSliver:
@kcristiano modified the title a while ago, hope this can be merged soon :)

kcristiano:
@totten would like to see if this can be merged.  Somewhere along the line we set loose permission in WP and never caught it.

kcristiano:
@eileenmcnaughton  Any chance you can take a look here.   It's been merge ready and would be good to get in.

JoeMurray:
@monish please review and merge

eileenmcnaughton:
Merging based on @kcristiano review and also because it seems more correct & consistent to me

PR #121, opened 2018-01-18
Installation - If available, use the new installer package

totten:
`civicrm-core` includes an installer (`civicrm/install/index.php`), but this
code is difficult to update/maintain.  The `civicrm-setup`
(http://github.com/civicrm/civicrm-setup/) package is a "leap" to replace it
with a more maintainable library.

This patch uses whichever is available.

Before
------
 * The page `civicrm-install` always uses `civicrm/install/index.php`.

After
-----
 * The page `civicrm-install` picks the first available installer UI, either:
   * `civicrm/vendor/civicrm/civicrm-setup/*` (`Civi\Setup::createController()`)
   * `civicrm/packages/civicrm-setup/*` (`Civi\Setup::createController()`)
   * `civicrm/setup/*` (`Civi\Setup::createController()`)
   * `civicrm/install/index.php`

Comments
-----------
 * If you actually want to try it out, just clone http://github.com/civicrm/civicrm-setup/ and put it in `civicrm/packages/civicrm-setup`.

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -805,6 +805,7 @@ class CiviCRM_For_WordPress {
         require_once $loader;
         require_once implode(DIRECTORY_SEPARATOR, [$civicrmCore, 'CRM', 'Core', 'ClassLoader.php']);
         CRM_Core_ClassLoader::singleton()->register();
+        \Civi\Setup::assertProtocolCompatibility(0.1);
         \Civi\Setup::init([
           'cms' => 'WordPress',
           'srcPath' => $civicrmCore,
--- a/civicrm.php
+++ b/civicrm.php
@@ -792,6 +792,36 @@ class CiviCRM_For_WordPress {
    * @return void
    */
   public function run_installer() {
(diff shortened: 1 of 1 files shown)


kcristiano:
@totten This looks great.  Just tested and it works as described.

WP 4.9.2 CiviCRM master.   Added patch and civicrm-setup to civicrm/packages

new installer is run, able to edit the DB credentials.   Upon entering incorrect credentials I get a failure notice.  Once corrected the installer is ready and installation proceeds and succeeds.

Is the plan to include civicrm-setup when we package the tarballs?

totten:
@kcristiano Exactly. We'll probably make two builds for WP -- the current one, and a new one with the updated installer.

kcristiano:
@totten I would love to see this in early in the RC cycle so we can get be testing as much as possible

totten:
Thanks @kcristiano. Agreed, I'd like to get this in soon, and I don't think there's much downside since (by default) it'll fallback to existing behavior.

PR #122, opened 2018-01-23
CRM-16421 - Installation - Update to newer protocol

totten:
This is a follow-up to the recently-merged #121. 

In #121, it was expected that some systems would opt-in to using the new installer, and others would opt-out. The mechanism for opt-in/opt-out was to download (or not download) the `civicrm-setup` folder.

This patch revises the protocol -- you can download the `civicrm-setup` folder regardless of preference. To indicate the preferred installer, create (or delete) a file named `.use-civicrm-setup`. In terms of the overall build/distribution infrastructure, it's easier to support this protocol.

---

 * [CRM-16421: Work to get CiviCRM for WordPress in WordPress' official Repository](https://issues.civicrm.org/jira/browse/CRM-16421)

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -801,15 +801,14 @@ class CiviCRM_For_WordPress {
     );
     foreach ($setupPaths as $setupPath) {
       $loader = implode(DIRECTORY_SEPARATOR, [$civicrmCore, $setupPath, 'civicrm-setup-autoload.php']);
-      if (file_exists($loader)) {
+      if (file_exists($civicrmCore . DIRECTORY_SEPARATOR . '.use-civicrm-setup') && file_exists($loader)) {
         require_once $loader;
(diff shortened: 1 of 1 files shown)


kcristiano:
Tested:

Built via distmaker adding the two related patches

created wp-empty and drupal empty via buildkit
Installed civicrm using *-drupal.tgz and *-wordpress.zip on these sites. Used existing installer, Install successfull

Installed on Joomla Test site, installed  CiviCRM using *-joomla.zip.  Used existing Installer, Install successful

created wp-empty via buildkit
Installed civicrm via *-wporg.zip
New Installer, successful install

@totten this looks good for merge

totten:
Coo. I know it's a bit tricky to test this stuff (what with distmaker and multiple patches). So thank you, @kcristiano !

PR #123, opened 2018-02-22
CRM-21788 Clean URLs (Proof of Concept)

christianwach:
This is one part of the required changes to implement Clean URLs (or at least parity with CiviCRM's URL schema in Drupal) in WordPress. Please refer to the Jira issue for accompanying changes to CiviCRM Core - or [visit the PR directly](https://github.com/civicrm/civicrm-core/pull/11708) since Jira doesn't seem to have picked it up.

Please note that I do not recommend this as a production-ready PR. I think there's going to be a lot of testing required since there are so many moving parts to this. Any takers? :-)

---

 * [CRM-21788: Clean URLs in WordPress](https://issues.civicrm.org/jira/browse/CRM-21788)

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -302,12 +302,6 @@ class CiviCRM_For_WordPress {
       session_start();
     }
 
-    // save original arrays
-    $this->civi_get     = $_GET;
-    $this->civi_post    = $_POST;
-    $this->civi_cookie  = $_COOKIE;
-    $this->civi_request = $_REQUEST;
-
     // get classes and instantiate
     $this->include_files();
 
@@ -1380,10 +1374,10 @@ class CiviCRM_For_WordPress {
     $this->wp_request = $_REQUEST;
 
     // reassign globals
-    $_GET     = stripslashes_deep($this->civi_get);
-    $_POST    = stripslashes_deep($this->civi_post);
-    $_COOKIE  = stripslashes_deep($this->civi_cookie);
-    $_REQUEST = stripslashes_deep($this->civi_request);
+    $_GET     = stripslashes_deep( $_GET );
+    $_POST    = stripslashes_deep( $_POST );
+    $_COOKIE  = stripslashes_deep( $_COOKIE );
+    $_REQUEST = stripslashes_deep( $_REQUEST );
 
     // test for query var
     $q = get_query_var( 'q' );
--- a/civicrm.php
+++ b/civicrm.php
@@ -545,7 +545,7 @@ class CiviCRM_For_WordPress {
     // register user hooks
     $this->users->register_hooks();
 
-	// have we flushed rewrite rules?
+    // have we flushed rewrite rules?
     if ( get_option( 'civicrm_rules_flushed' ) !== 'true' ) {
 
       // apply custom rewrite rules, then flush rules afterwards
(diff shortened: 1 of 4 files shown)


kcristiano:
@christianwach Fantastic work.    I've built test sites with these patches and they work as expected.

I've run upgrades from 4.7.27 to latest master as well as fresh installs.

The code looks  good and results are as expected.

We plan on further testing as this will need a good going through to find any issues.  I'd recommend merging this PR and https://github.com/civicrm/civicrm-core/pull/11708 early to get in the next RC.

mlutfy:
Can someone test this with Mosaico?

The patch seems to work really well, but when testing Mosaico, CiviMail could not save drafts anymore. It would respond with "Unable to decode supplied JSON". I got lost debugging the json before/after, so before continuing further, I just wanted to make sure it was not a red herring.

(I did try to apply the patch, run tests, stash, run tests again, etc, every time with the same result)

civicrm.php
```diff
@@ -1245,10 +1380,45 @@ public function remove_wp_magic_quotes() {
...
+
+		if (!empty($action)) { $_REQUEST['action'] = $_GET['action'] = $action; }
+		if (!empty($mode)) { $_REQUEST['mode'] = $_GET['mode'] = $mode; }
+		if (!empty($cid)) { $_REQUEST['cid'] = $_GET['cid'] = $cid; }
+		if (!empty($gid)) { $_REQUEST['gid'] = $_GET['gid'] = $gid; }
+		if (!empty($sid)) { $_REQUEST['sid'] = $_GET['sid'] = $sid; }
+		if (!empty($cs)) { $_REQUEST['cs'] = $_GET['cs'] = $cs; }
+		if (!empty($force)) { $_REQUEST['force'] = $_GET['force'] = $force; }
+
+    }
```
mlutfy: Syntax: should be 2 spaces, not 2 tabs?

christianwach: Gah. Editor defaults to tab-indent. Will rectify.

christianwach:
@mlutfy Thanks for testing.

> reverting these lines fixes it

Cheers, that's a help. I'll test this with Mosaico.

christianwach:
@mlutfy When you say:

> reverting these lines fixes it

Do you mean "removing these lines"?

mlutfy:
I used:

```
$_POST    = stripslashes_deep($_POST);
```

Instead of:

```
$_POST    = stripslashes_deep($this->civi_post);
```

Line 1249: https://github.com/civicrm/civicrm-wordpress/pull/123/files#diff-d1bc6ad2b4d05fe1a841fce94b11b8dbL1248

christianwach:
@mlutfy Thanks for clarifying.

mlutfy:
I stumbled on this bug:

* Create a public CiviEvent
* Try to register to that event (ex URL produced by CiviCRM: https://example.org/civicrm/event/register/?id=285&reset=1)

The registration works, and is saved in CiviCRM, however, the redirection seems to fail to show the ThankYou page:

https://example.org/civicrm/event/register/?_qf_ThankYou_display=true&qfKey=5b22c771b131249a1b40ab7ee6e7aee4_1884

I get a fatal error with: "Could not find valid value for id".

I tested with and without my revert of `$_POST    = stripslashes_deep($_POST);`.

christianwach:
@mlutfy Thanks for reporting. Just to be sure - you definitely have [this commit in core](https://github.com/civicrm/civicrm-core/pull/11708/commits/cea37e01340bebebe3aca730d93bfd83668f7b2c)? That was causing the same problem for me by triggering a browser redirect on account of `?&` in the initial URL. If so, then there must be other places in core that cause a redirect after the session has been cleared.

kcristiano:
@christianwach Do we want to close this and test off of your fork?    It may be better to track as an issue until you feel it is ready to merge.  I'd also like to see https://github.com/civicrm/civicrm-wordpress/pull/125 in sooner rather than laster and that will require this to be rebased

cc @mlutfy

PR #125, opened 2018-02-28
CRM-21812 - don't specify installation type post install

MegaphoneJon:
The wp-cli-login-server plugin checks for the presence of elements in the $_GET array, and fails if any are found.  This conflicts with CiviCRM, which places:

```
$_GET['civicrm_install_type'] = 'wordpress';
```

into every page load.

I confirmed that this value is only used for install, and placed a conditional around it to only insert pre-install.  I successfully installed CiviCRM and engaged in normal use with the conditional applied, and confirmed that this fixed the WP plugin.

This is intended to close https://github.com/aaemnnosttv/wp-cli-login-command/issues/24.

---

 * [CRM-21812: WordPress install variable causes conflict with some WP plugins](https://issues.civicrm.org/jira/browse/CRM-21812)

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -308,7 +308,9 @@ class CiviCRM_For_WordPress {
     if ( $this->civicrm_in_wordpress() ) {
       // this is required for AJAX calls in WordPress admin
       $_GET['noheader'] = TRUE;
-    } else {
+    }
+
+    if ( !CIVICRM_INSTALLED && !$this->civicrm_in_wordpress() ) {
       $_GET['civicrm_install_type'] = 'wordpress';
     }
 


christianwach:
FWIW, this is also fixed in #123 as can be seen [here](https://github.com/christianwach/civicrm-wordpress/blob/821e78f1c2236d17cdee6214575301bb5d987dee/civicrm.php#L663-L664).

kcristiano:
I prefer https://github.com/civicrm/civicrm-wordpress/pull/123 in favor of this PR.   They solve the same issue, but #123 has other benefits as well.

kcristiano:
@christianwach  as https://github.com/civicrm/civicrm-wordpress/pull/123 needs more testing I think we should merge this now and rebase #123 Any objections?

christianwach:
@kcristiano No objections here

MegaphoneJon:
@eileenmcnaughton This has been reviewed by @christianwach and @kcristiano, but I don't think any of us have merge rights.  Are you able to merge this for us?  Thanks!

eileenmcnaughton:
Yes

PR #126, opened 2018-03-26
Gutenberg Compatibility - Do not render shortcodes in REST context

bastienho:
which breaks saving in Gutenberg editor.

Symptom:

**Notice**:  Trying to get property of non-object in `civicrm/civicrm/CRM/Utils/System/WordPress.php` on line **221**```

### DIFF
--- a/includes/civicrm.shortcodes.php
+++ b/includes/civicrm.shortcodes.php
@@ -246,7 +246,11 @@ class CiviCRM_For_WordPress_Shortcodes {
    * @return string HTML for output
    */
   public function render_single( $atts ) {
-
+    // Do not parse shortcodes in REST context, which breaks saving in Gutenberg editor
+    if(defined('REST_REQUEST') && REST_REQUEST){
+        return;
+    }
+   
     // check if we've already parsed this shortcode
     global $post;
     if ( is_object($post) ) {


kcristiano:
@bastienho Thank you for the PR.    

Can you detail the steps to reproduce?   I imagine you are inserting a shortcode with Gutenberg.   Can you also let us know the version of Gutenberg that you are using it will be helpful for testing.

bastienho:
Hi,

this can be reproduced in WordPress 4.9.4 with Gutenberg 2.4.0 (which will become the new core editor).

BR

tadcol:
Thanks @bastienho As Gutenberg is constantly cahnging (from a plugin perspective) the version is good to know.

Can you tell me what type of block you inserted the shortcode in?  Was it the shoirtcode block?  The standard Content (Paragraph) block?   The Classic Block?  

I want to reproduce and test the patch both with and without Gutenberg.

kcristiano:
Thanks @bastienho As Gutenberg is constantly cahnging (from a plugin perspective) the version is good to know.

Can you tell me what type of block you inserted the shortcode in? Was it the shoirtcode block? The standard Content (Paragraph) block? The Classic Block?

I want to reproduce and test the patch both with and without Gutenberg.

bastienho:
it is the shortcode block

kcristiano:
[Core Issue 33](https://lab.civicrm.org/dev/core/issues/33)

Steps to reproduce:

- Install WP 4.9.4 and CiviCRM 5.0beta1
- Install Gutenberg plugin for WP https://wordpress.org/plugins/gutenberg/
- Add a post or page with Gutenberg
- Copy a CiviCRM shortcode in the shortcode block ```[civicrm component="contribution" id="1" mode="test"]```
- Publish the post

You will get an on screen error that editing failed and the logs will show:
`Notice: Trying to get property of non-object in civicrm/civicrm/CRM/Utils/System/WordPress.php on line 221`

Apply the patch.
Repeat the steps and Publishing is successful

I also tested on CiviCRM 5.0.beta1 with the patch and without the Gutenberg plugin.   Publishing posts with the shortcode works properly.

We should merge this in to 5.0beta1 as soon as possible and target this patch for the next release.

cc: @christianwach @totten @colemanw

totten:
Merging based on @kcristiano's review. Thanks, @bastienho @kcristiano.

totten:
@bastienho @kcristiano @christianwach - Just double-checking a bit... this patch purports to disable shortcodes in Gutenberg, but it looks like it actually disables Civi shortcodes for *all* processing via WP's REST end-point?

I'm no expert in WP's ecosystem, so there's some important context I'm missing. *What exactly do people use WP's REST API for?* If they *only* use it for Gutenberg, then great. If they use it for other things (like fetching WP pages and showing the content in another site), then wouldn't his break short-code handling for others?

kcristiano:
@totten the WP REST API is used for many things,I have not come across any other issues.  You make a good point and we can do some testing.

christianwach:
For my part (I haven't had time to test as yet) I would first be curious why `global $post` is empty [in core](https://github.com/civicrm/civicrm-core/blob/master/CRM/Utils/System/WordPress.php#L221). Like @totten I have reservations about the impact of this patch outside its intended purpose.

totten:
OK, since Christian had the same concern, I'm inclined to let this stew a little longer -- rather than do it as a last-minute addition to `5.0`, let's give a full cycle and move it to`5.1`.

bastienho:
An answer to @totten 's question, more about REST in WordPress than about Gutenberg:

In PUT context, the WP REST API returns both raw and rendered content.

In fact, the error occurs in `CRM_Core_Controller` which does not seems to deal with this context (Gutenberg or not), a valid key is required, and there is no way to pass the `ignoreKey` parameter, even by adding `page_arguments[ignoreKey]=true` in request args.

At this point we have two solutions:
- Patch the core to allow `ignoreKey` (in *CRM_Core_Menu*, I imagin)
- Do not render the shortcode in REST / PUT context 

Hope it helps

christianwach:
@bastienho Thanks for this - seems we came to [similar conclusions](https://github.com/civicrm/civicrm-wordpress/pull/129#issuecomment-379227990) :-)

PR #129, opened 2018-04-04
Gutenberg Compatibility - Do not render shortcodes in REST context

totten:
Overview
This prevents hard errors in Gutenberg by disabling short-code rendering in any REST calls.

See also: https://lab.civicrm.org/dev/wordpress/issues/1

This PR is a resubmission of @bastienho's #126, which was reverted to give more time for consideration.

### DIFF
--- a/includes/civicrm.shortcodes.php
+++ b/includes/civicrm.shortcodes.php
@@ -246,7 +246,11 @@ class CiviCRM_For_WordPress_Shortcodes {
    * @return string HTML for output
    */
   public function render_single( $atts ) {
-
+    // Do not parse shortcodes in REST context, which breaks saving in Gutenberg editor
+    if(defined('REST_REQUEST') && REST_REQUEST){
+        return;
+    }
+   
     // check if we've already parsed this shortcode
     global $post;
     if ( is_object($post) ) {


totten:
@bastienho @kcristiano @christianwach - Continuing the previous review discussion, if we still wanted the quick mitigation... could one do a similar patch that keys off of Gutenberg more directly? Maybe one of these?

 * "if gutenberg is enabled, then don't render shortcodes"
 * "if this is rest call and gutenberg is enabled, then don't render shortcodes"

bastienho:
You're right. let's do that.

christianwach:
I had a quick look at this and it seems that the REST calls made by Gutenberg happen such that `is_admin()` evaluates to `FALSE`. They also seem to happen in a way that CiviCRM cannot connect the request to authentication tokens and/or the user session. This is partly what throws errors in CiviCRM for me (leaving aside the error in `CRM_Utils_System_WordPress::url()` for now - which FWIW my clean-urls PR seems not to be affected by) but is supplemented by the fact that whilst Civi thinks this is a front-end request, the `wp` action has not occurred - thus skipping the shortcode `prerender()` method.

What seems relevant to me here is not that we disable shortcode parsing whilst in a `REST_REQUEST` context, but that that they do not get rendered during `POST`, `DELETE` or `PUT` requests. I can definitely see a use case for rendering the shortcodes via `REST_REQUEST` and `GET` (for retrieving and rendering a page/post, for example) but can't see one in any other case. Perhaps this helps refine when the shortcode rendering happens and when not?

totten:
OK, glad there's agreements coming up.

Trying to get a 'take away' for me (as a general facilitator), it sounds like the patch needs more work, so my cherry-picked/forward-ported PR isn't getting merged in its current form. Let me close it and encourage @bastienho or @christianwach to open a better PR.  :)

PR #131, opened 2018-05-23
[WIP][DO NOT MERGE] Allowed user defined shortcodes to be rendered

Edzelopez:
Overview
This PR will add support to render user defined shortcodes in WordPress when specified through extensions.

Before
Code returns abruptly with an error saying "Do not know how to handle this shortcode"

After
Code renders the shortcode based on filters specified in the extension.

Technical Details
_If the PR introduces noteworthy technical changes, please describe them here. Provide code snippets if necessary_

Comments
_Anything else you would like the reviewer to note_

### DIFF
--- a/includes/civicrm.shortcodes.php
+++ b/includes/civicrm.shortcodes.php
@@ -649,9 +649,14 @@ class CiviCRM_For_WordPress_Shortcodes {
         break;
 
       default:
-
-        echo '<p>' . __( 'Do not know how to handle this shortcode', 'civicrm' ) . '</p>';
-        return;
+        $args = apply_filters( 'civicrm_shortcode_preprocess_atts', $args, $shortcode_atts );
+        if (!$args) {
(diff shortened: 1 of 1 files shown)


kcristiano:
@christianwach Can you take a look at this?

JoeMurray:
@Edzelopez please update the overview. Also could you or @monish respond to @christianwach promptly?

Edzelopez:
Thanks for looking at this @christianwach and apologies for the delay. You are right, I intend to allow the use of user defined shortcodes before the control is broken by the switch case. I will modify the code to allow logic to flow to the filter rather than declaring it a second time above.

christianwach:
@Edzelopez That would be great :)

christianwach:
This can be closed in favour of #134 which implements code based on the advice that I gave in [my comment on this PR](https://github.com/civicrm/civicrm-wordpress/pull/131#issuecomment-404189434).

kcristiano:
closing, #134 completes the work suggested

PR #136, opened 2018-10-03
Update profile shortcode to use new front-end title

bhahumanists:
Overview
Civi profiles have a new option to show a different title on the front end. This updates the shortcode to use this title if it's available.

Before
Profile uses the title field.

After
Profile uses frontend_title field if available, and falls back to title if not.

### DIFF
--- a/includes/civicrm.shortcodes.php
+++ b/includes/civicrm.shortcodes.php
@@ -793,7 +793,11 @@ class CiviCRM_For_WordPress_Shortcodes {
         $civi_entity = civicrm_api( 'uf_group', 'getsingle', $params );
 
         // set title
-        $data['title'] = $civi_entity['title'];
+        if ( isset($civi_entity['frontend_title']) ) {
+          $data['title'] = $civi_entity['frontend_title'];
+        } else {
+          $data['title'] = $civi_entity['title'];
+        }
 
         // set text to empty
         $data['text'] = '';


christianwach:
@kcristiano Looks good to me.

kcristiano:
@bhahumanists This works, but only when multiple shortcodes are used.   We'll need to dig deeper to allow changing the title on Single pages.  CRM_Utils_System_WordPress is setting the title for Single Pages.

Do you want to take a look at that and add to this PR?

bhahumanists:
Oops - missed that. Yes, I'll update. Thanks for the heads up.

christianwach:
@bhahumanists I'm not sure you'll be able to complete this task in this plugin. I reckon it's going to need changes to Core because when I view a profile directly via the basepage e.g.

`http://my.civicrm.local/civicrm/?page=CiviCRM&q=civicrm/profile&gid=1&reset=1` 

I do not see the `frontend_title` implemented. 

Looks like a number of files need updating: e.g. [here](https://github.com/civicrm/civicrm-core/blob/master/CRM/Profile/Form/Edit.php#L165), [here](https://github.com/civicrm/civicrm-core/blob/master/CRM/Profile/Page/Dynamic.php#L357) and [here](https://github.com/civicrm/civicrm-core/blob/master/CRM/Profile/Page/Listings.php#L307). Probably more need changing, but these stood out.

christianwach:
@kcristiano Unless I'm mistaken, I'd say this PR can be merged since there's no more that can be done here.

kcristiano:
Since this will only allow for displaying `frontend_title` when multiple shortcodes are used, I am concerned that the the Title would be displayed differently to the user depending on the page they were on.   The title could be 'New fancy Title' on a page with Multiple shortcodes, but upon clicking the Title link the single page would have a different title.  I think this will lead to confusion.  

I would think it better to address the issue in civicrm-core as detailed here  https://github.com/civicrm/civicrm-wordpress/pull/136#issuecomment-427374925 before Merging.  

@bhahumanists @christianwach what are your thoughts?

PR #143, opened 2018-12-17
Fix fatal error with some versions of PHP

christianwach:
Overview
Simple fix to prevent fatal error with some versions of PHP. Fixes [Issue 16](https://lab.civicrm.org/dev/wordpress/issues/16).

Before
Some users report

```
Fatal error: Can't use function return value in write context in .../public_html/CRM/wp-content/plugins/civicrm/civicrm.php on line 342
```

After
Should not throw fatal error.


eileenmcnaughton:
@kcristiano good to merge?

PR #147, opened 2019-02-13
CMS path fix in wp-cli

aydun:
Overview
The path specified to wp-cli as `--path` is not passed to CiviCRM resulting in URL errors.

See dev/wordpress#18

Before
The path specified by `wp-cli --path` is not passed to CiviCRM.  One symptom of this is that mails delivered by `wp-cli --path XX --url http://XX api Job.execute` with click-through tracking enabled can have bad URL's.  

Steps to reproduce error:
- create a mailing including links, check 'click-through tracking' is enabled, submit mail
- from command line:
  - cd to somewhere _outside_ of the wp tree
  - run `wp-cli --user=cron --path=XX --url=http://YY api Job.execute auth=0`
- check mail - links don't work

After
The path specified by `wp-cli --path` is passed to CiviCRM.  Mails have correct trackable URL's

### DIFF
--- a/wp-cli/civicrm.php
+++ b/wp-cli/civicrm.php
@@ -165,6 +165,13 @@ if ( ! defined( 'CIVICRM_WPCLI_LOADED' ) ) {
         return WP_CLI::error( "Unrecognized command - '$command'" );
       }
 
+      # if --path is set, save for later use by Civi
+      global $civicrm_paths;
+      $wp_cli_config = WP_CLI::get_config();
+      if (!empty($wp_cli_config['path'])) {
+        $civicrm_paths['cms.root']['path'] = $wp_cli_config['path'];
+      }
+
       # run command
       return $this->{$command_router[ $command ]}();
 


kcristiano:
@aydun Thanks for the PR.  

Can you separate out the code formatting changes into another PR?  I don't want this change to be lost amongst all the whites pace cleanup.

aydun:
@kcristiano - sure, see #148

kcristiano:
* ([`r-explain`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-explain)) Pass
* ([`r-test`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-test)) NA
* ([`r-code`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-code)) This PR still contains the whitespace changes.  I would prefer to have this only be the code change and we can look at the NFC separately. 
* ([`r-doc`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-doc)) NA
* ([`r-run`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-run)) I ran a mosaico mailing with and without the patch.  I could not reproduce the error.   However the code makes sense and this is an improvement.   Mailings and other wp-cli jobs work as expected with the patch.   


@aydun Thanks for the PR I have tested and would like to merge.  Can you remove the NFC commit from your local branch so that the only commit here is the fix for the issue?   Once that is done we can merge.

aydun:
Thanks for testing @kcristiano   Could you merge #148 then I can rebase it out of this.  I've updated the description with more detail for reproducing the error.

kcristiano:
@aydun I am actually not convinced we need to do the white space change.   Are you converting to CiviCRM code style or WP.

In an earlier commit the file was changed to CiviCRM code style.

aydun:
@kcristiano following the discussion around #146 the updates there are just doing the tab->space conversion.  I have not converted fully to CiviCRM code style as it seems there is not agreement yet on whether that is desirable - but I understood that discussion to mean there was agreement at least on the tab/space issue.

kcristiano:
Very valid point @aydun I'll look at that PR separate from this one.     I do want to be cautious that we don;t have any united changes in a NFC formatting change.  That last code style change I referred to resulted in some wp-cli breakage.

kcristiano:
@aydun Merged 148 so we can merge once you remove the NFC commit

aydun:
@kcristiano rebased now

PR #151, opened 2019-04-25
Fix args when called via AJAX

christianwach:
Overview
Duplicate of #150 for 5.13 branch

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -1566,8 +1566,8 @@ class CiviCRM_For_WordPress {
 
     // Fix 'civicrm/civicrm' elements derived from CRM:url()
     // @see https://lab.civicrm.org/dev/rc/issues/5#note_16205
-    if (stristr($q, 'civicrm/civicrm') !== false) {
-      $q = str_replace('civicrm/civicrm', 'civicrm', $q);
+    if (substr($q, 0, 16) === 'civicrm/civicrm/') {
+      $q = str_replace('civicrm/civicrm/', 'civicrm/', $q);
       $_REQUEST['q'] = $_GET['q'] = $q;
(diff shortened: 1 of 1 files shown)


kcristiano:
See https://github.com/civicrm/civicrm-wordpress/pull/150#issuecomment-486370353

civicrm.php
```diff
@@ -1564,6 +1564,14 @@ public function get_request_args() {
       $q = isset($_GET['q']) ? $_GET['q'] : '';
     }
 
+    // Fix 'civicrm/civicrm' elements derived from CRM:url()
```
kcristiano: @christianwach Would it makes sense to conditionally load this only if `CIVICRM_CLEANURL` is True?

christianwach: @kcristiano Hmm, good point. Though I can't think of a situation where you'd get a URL of the form:

```
http://domain.tld/civicrm/?page=CiviCRM&q=civicrm%2Fcivicrm%2Fsomething&reset=1&id=1
```

I can't find any instances of this.

Also, I'll have to check if `get_request_args()` is ever called before `CIVICRM_CLEANURL` has been defined.

kcristiano: @christianwach I am being overly cautious.   The PR works and should be merged, I was curious if we could tighten it up further.

kcristiano:
Tested with the updated patch and this corrects the issue.

civicrm.php
christianwach: @kcristiano I've done some testing and cannot find an instance where `get_request_args()` is called before `CIVICRM_CLEANURL` has been defined... but I agree that it's always best to be cautious. Hence I've created #153 and #154

PR #153, opened 2019-04-26
Only check path when using Clean URLs

christianwach:
Overview
Further to #150 and #151 and https://github.com/civicrm/civicrm-wordpress/pull/151#discussion_r278666038, this trivial change fixes Javascript-derived malformed paths only when `CIVICRM_CLEANURL` is defined and switched on.


kcristiano:
will merge 5.13 branch to master.  Closing

PR #154, opened 2019-04-26
Only check path when using Clean URLs (5.13)

christianwach:
Overview
Duplicate of #153 against 5.13 branch


kcristiano:
Tested on 5.13 with and without CleanURLs enable.

Issue fixed. 

This improves the original PR.   Merging

PR #156, opened 2019-06-04
dev/wordpress#30 revert how we get query var

lcdservices:
Overview
At the point where the WP function get_query_var(), the page vars may not have been loaded, causing Civi frontend pages to load with just the WP page content ("don't delete this page...").

Before
Civi frontend pages don't load properly.

After
Civi frontend pages load properly.

Technical Details
This changes uses the Civi Request class to retrieve the page var instead of get_query_var().

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -384,11 +384,8 @@ class CiviCRM_For_WordPress {
    */
   public function civicrm_in_wordpress_set() {
 
-    // Get identifying query var
-    $page = get_query_var( 'page' );
-
     // Store
-    self::$in_wordpress = ( $page == 'CiviCRM' ) ? TRUE : FALSE;
+    self::$in_wordpress = (isset($_GET['page']) && $_GET['page'] == 'CiviCRM') ? TRUE : FALSE;
 
   }
 


christianwach:
@lcdservices Thanks for highlighting the issue. I'm copying in some of our discussion for context.

tl;dr I'd prefer to revert to the previous code:

```php
self::$in_wordpress = ( isset( $_GET['page'] ) && $_GET['page'] == 'CiviCRM' ) ? TRUE : FALSE;
```

After digging a bit, I agree that `get_query_var()` is not the appropriate way to test for $_GET['page'] at that particular point in the load process. It seems that `get_query_var()` is only useful once the `parse_request` action has fired and `WP_Query` is populated. The `civicrm_in_wordpress_set()` method is called during the `init` action, which precedes `parse_request` and can therefore return incorrect values.

[Aside: in a wider context, it's my opinion that `civicrm_in_wordpress_set()` is a pointless method -- where else would CiviCRM be? -- and why is it false when it's a Shortcode? -- and if that's all it detects, then it's poorly named and the plugin needs refactoring to get rid of it.

FWIW, I'd prefer to revert to the previous known-good code. I see no problem with querying the `$_GET` superglobal at that point and returning a boolean based on that query -- it's not like the original code does anything with the contents of `$_GET['page']` after all.

lcdservices:
@christianwach I've reverted to the original code, as discussed.

christianwach:
@lcdservices This doesn't look right to me as it stands: I thought the idea was to revert to [the method as it was written previously](https://github.com/civicrm/civicrm-wordpress/blob/9ccb2ecdf82b85a946716aaea047512f12867c79/civicrm.php#L330-L347)?

lcdservices:
ok... I mean, it's the same thing. I just left the structure as you had previously modified it because I find it more readable. But I can fully revert if you prefer.

lcdservices:
actually, I take it back. I screwed it up with an ugly mix of both. my bad. will fix in a minute.

lcdservices:
@christianwach updated again

christianwach:
@kcristiano @lcdservices I've been looking into this a bit further. It makes no sense to revert back to the previous code because it won't work when Clean URLs are enabled. Let me think this through a bit more and I'll post my findings here.

christianwach:
So then, here are the results of my investigation:

##### On the `civicrm_in_wordpress_set()` method:

* The only place that actually calls `civicrm_in_wordpress_set()` that I can find is in [the `register_hooks_front_end()` method](https://github.com/civicrm/civicrm-wordpress/blob/942e3a7527b27d6dcac0b900093726cd83dba579/civicrm.php#L558).
* @lcdservices I was wrong about the `civicrm_in_wordpress_set()` method being called during the `init` action -- I had forgotten that `register_hooks_front_end` is the [callback for the `parse_query` action](https://github.com/civicrm/civicrm-wordpress/blob/942e3a7527b27d6dcac0b900093726cd83dba579/civicrm.php#L538), which fires *after* the `parse_request` action.
* Consequently `get_query_var()` *should* return the correct value -- and in my testing *always does*. 
* Here are [the docs on `parse_query`](https://codex.wordpress.org/Plugin_API/Action_Reference/parse_query) which I must have read at the time that I last refactored the plugin (and had forgotten about when I first looked at this issue) since that's precisely why I delay everything else until query has been parsed.

##### On the legacy `civicrm_wp_in_civicrm()` function which calls `civi_wp()->civicrm_in_wordpress()`:

* The docblock [states](https://github.com/civicrm/civicrm-wordpress/blob/942e3a7527b27d6dcac0b900093726cd83dba579/civicrm.php#L1782) that it is "Called by `setTitle()` in civicrm/CRM/Utils/System/WordPress.php" but this is [no longer true](https://github.com/civicrm/civicrm-core/blob/31008c053feaa652635cd2d946e7421263ae8775/CRM/Utils/System/WordPress.php#L66-L67).
* I have searched GitHub and have found no contemporary usage of `civicrm_wp_in_civicrm()`.
* I think we can therefore conclude that this can safely be ignored.

##### This leaves the `civicrm_in_wordpress()` method:

* There only seem to be two places that call `civicrm_in_wordpress()` that I can find: 
  1. [the `register_hooks_front_end()` method](https://github.com/civicrm/civicrm-wordpress/blob/942e3a7527b27d6dcac0b900093726cd83dba579/civicrm.php#L561)
  2. [the `invoke()` method](https://github.com/civicrm/civicrm-wordpress/blob/942e3a7527b27d6dcac0b900093726cd83dba579/civicrm.php#L1363).
* From this, we can see that the method is used for only two things:
  1. It determines if the hooks should be registered for a Shortcode (or another front-end CiviCRM "content type") in `register_hooks_front_end()`
  2. If `$_REQUEST['noheader']` and `$_GET['noheader']` should be set in `invoke()`.

I stand by what I wrote elsewhere: that "it's my opinion that `civicrm_in_wordpress_set()` is a pointless method -- where else would CiviCRM be? -- and why is it false when it's a Shortcode? And if that's all it detects, then it's poorly named and the plugin needs refactoring to get rid of it." In the light of the above analysis, I suggest that it could be renamed to describe what it does.

@lcdservices I think you need to look deeper as to why `get_query_var()` is failing on your client's site. Perhaps another plugin is overwriting `$_GET['page']`? Or perhaps something is filtering the query variables?

@kcristiano In conclusion, I think this PR should not be merged since the code works as it stands -- whether Clean URLs are enabled or disabled.

PR #157, opened 2019-06-06
Recreate rewrite rules when basepage setting is updated

christianwach:
Overview
When Clean URLs are enabled, it can sometimes be difficult to reliably flush the rewrite rules such that Clean URLs are applied. This PR causes those rules to be flushed when saving the WordPress Base Page setting in CiviCRM.

Before
Rewrite rules were not flushed when saving the WordPress Base Page setting in CiviCRM.

To reproduce:

1. Create a new WordPress page, for example `https://domain.org/my-new-basepage`
2. In CiviCRM, navigate to "Administer" / "System Settings" / "CMS Database Integration"
3. Change the base page slug to `my-new-basepage` and hit "Save"
4. Navigate to, for example, "Contributions" / "Manage Contribution Pages" and click "Links" / "Live Page" for one of the pages. The result will be a 404.

After
Rewrite rules are flushed when saving the WordPress Base Page setting in CiviCRM.

To test:

Same procedure as before, but this time, clicking the "Live Page" link shows the populated Contribution Page.

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -676,6 +676,9 @@ class CiviCRM_For_WordPress {
     // Print CiviCRM's header
     add_action('admin_head', array( $this, 'wp_head' ), 50);
 
+    // Listen for changes to the basepage setting
+    add_action( 'civicrm_postSave_civicrm_setting', array( $this, 'settings_change' ), 10 );
+
     // If settings file does not exist, show notice with link to installer
     if ( ! CIVICRM_INSTALLED ) {
       if ( isset( $_GET['page'] ) && $_GET['page'] == 'civicrm-install' ) {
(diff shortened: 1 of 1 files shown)


kcristiano:
@christianwach We should document the process in the Administrator manual before merge.  I had to find the prior merged PR to get the correct settings

kcristiano:
@christianwach The code looks sensible, in testing it works fine on single site and Multi-Site.

I cannot reproduce the issue where flushing permalinks failed.



* ([`r-explain`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-explain)) Pass
* ([`r-test`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-test)) Pass
* ([`r-code`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-code)) Pass
* ([`r-doc`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-doc)) Needs Docs
* ([`r-run`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-run)) Pass: Installed fresh Single site, enabled Cleaner Urls, changed base page, flush worked.  Using existing Multisite enabled cleaner Urls was able to flush cache on all sub sites

I would like to see this in sooner rather than later for testing.   I do want to be sure we get admin docs done before 5.16 (which this should target).

Any thoughts on how to reproduce the error?   We can look at the documentation at the Sprint next t week  

cc @danaskallman

christianwach:
@kcristiano Steps to reproduce are in the original description.

I'm not convinced that docs are necessary since this change will be invisible to users. All it does is implement expected behaviour.

kcristiano:
@christianwach I can get the issue to reproduce on a single site, but I could not get it to reproduce on my local multi-site.  This could be an issue on my side.

I don't see a downside to this.

My comment on documentation refers to Cleaner URLs in WP, not this patch.   

As 5.15 was just released do we want this to be in the current RC?

christianwach:
> I could not get it to reproduce on my local multi-site

@kcristiano That reminds me - I've got to look at base pages and multisite installs more closely. Would be good to get #158 merged ASAP so I can move on that.

> My comment on documentation refers to Cleaner URLs in WP, not this patch.

Ah, okay, understood.

> As 5.15 was just released do we want this to be in the current RC?

You know far more about the release cycles than I do so I'm happy to go with whatever you suggest.

PR #159, opened 2019-06-14
Add WP-oriented E2E test suite, with HookTest as an example

totten:
Overview
Adds an empty test suite with one example test.

Before
* There are no E2E tests specific to WordPress.

After
* There is one E2E test specific to WordPress. It checks that integration works between CiviCRM hooks and WP actions/filters.

Technical Details
To run this in a `civibuild` site:

```
$ cd build/wpmaster/wp-content/plugins/civicrm
$ phpunit5
```

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -1,0 +1,24 @@
+<phpunit backupGlobals="false"
+         backupStaticAttributes="false"
+         colors="true"
+         convertErrorsToExceptions="true"
+         convertNoticesToExceptions="true"
+         convertWarningsToExceptions="true"
+         processIsolation="false"
+         stopOnFailure="false"
+         syntaxCheck="false"
+         beStrictAboutTestsThatDoNotTestAnything="false"
+         bootstrap="tests/phpunit/bootstrap.php"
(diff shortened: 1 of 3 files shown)


kcristiano:
@christianwach I worked on testing at the sprint.  This will enable us to have WP specific Unit Tests.   They are not required now, but we may eventually start requiring them.

totten:
jenkins, test this please

totten:
> They are not required now, but we may eventually start requiring them.

Tangentially... this PR creates a new test suite, `CiviWP`, which could do tests for functionality specifically in `civicrm-wordpress`. Separately, there is an existing test-suite called [E2E](https://github.com/civicrm/civicrm-core/tree/master/tests/phpunit/E2E) for functionality that is not specifically about one CMS but which needs some CMS to run.

I've enabled testing for `E2E` on `civicrm-wordpress` PRs. But, as @kcristiano says, the `CiviWP` is not required now. (Since they're not actually in the codebase, it doesn't make sense to require them. 😜 )

kcristiano:
Tested and works

PR #161, opened 2019-08-20
(dev/drupal#79) Fail more gracefully when upgrading on PHP 5.x

totten:
This version requirement officially went up to PHP 7.0 circa Civi 5.14. However, at that time, the upgrade metadata was kept at PHP 5.6 to allow somewhat softer landing for stragglers.  That's no longer possible in Civi 5.16+,

This just gives a clearer error when someone tries to upgrade with PHP 5.x.

Before
------

When accessing a Civi-WordPress site on PHP 5.6 (or running wp-cli), the Civi class-loader fails to initialize.

```
Parse error: syntax error, unexpected ':', expecting '{'
in /Users/totten/bknix/build/wpmaster/web/wp-content/plugins/civicrm/civicrm/vendor/league/csv/src/functions.php on line 33
```

(Approximate call-path: `civicrm.php` => `civicrm.settings.php` =>`CRM_Core_ClassLoader` => `vendor/autoload.php` => `vendor/league/csv/src/functions.php`)

After
-----

When accessing a Civi-WordPress site on PHP 5.6 (or running wp-cli), the error message is:

```
CiviCRM requires PHP version 7.0.0 or greater. You are running PHP version 5.6.38
```

Comments
--------

The canonical representation of the minimum PHP version is in `$civicrm_root/CRM/Upgrade/Form.php`.  However, correctly reading that metadata requires loading `civicrm.settings.php`, which triggers the crash.

To work around this, this patch reproduces the constant and uses a unit-test to ensure its continued accuracy.

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -832,6 +832,19 @@ class CiviCRM_For_WordPress {
   // CiviCRM Initialisation
   // ---------------------------------------------------------------------------
 
+  protected function assertPhpSupport() {
+    // Need to check this before bootstrapping - once we start bootstrapping, the error messages will become ugly.
+    if ( version_compare( PHP_VERSION, CIVICRM_WP_PHP_MINIMUM ) < 0 ) {
+      echo '<p>' .
+         sprintf(
+          __( 'CiviCRM requires PHP version %s or greater. You are running PHP version %s', 'civicrm' ),
+          CIVICRM_WP_PHP_MINIMUM,
+          PHP_VERSION
+         ) .
+         '<p>';
+      exit();
+    }
+  }
 
   /**
    * Initialize CiviCRM.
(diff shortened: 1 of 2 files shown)


kcristiano:
Thanks @totten This looks good to me

kcristiano:
This PR and the corresponding Core PR https://github.com/civicrm/civicrm-core/pull/15082  prevent upgrade.

However a fresh install fails with a 500 error 

```
An error of type E_PARSE was caused in line 33 of the file /home/civicrm/public_html/wp-content/plugins/civicrm/civicrm/vendor/league/csv/src/functions.php. Error message: syntax error, unexpected ':', expecting '{'
```

totten:
> However a fresh install still fails with a 500 error

Good point, @kcristiano. I moved the pre-boot PHP version check into a small helper function and updated so that both upgrade+install use-cases hit it.

kcristiano:
Thanks @totten Retested and works.

kcristiano:
@totten do we need to backport to 5.16?

totten:
@kcristiano yup

PR #162, opened 2019-08-21
Fail more gracefully when upgrading on PHP 5.x  - 5.16 backport

kcristiano:
Backport https://github.com/civicrm/civicrm-wordpress/pull/161

This version requirement officially went up to PHP 7.0 circa Civi 5.14. However, at that time, the upgrade metadata was kept at PHP 5.6 to allow somewhat softer landing for stragglers. That's no longer possible in Civi 5.16+,

This just gives a clearer error when someone tries to upgrade with PHP 5.x.

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -832,6 +832,19 @@ class CiviCRM_For_WordPress {
   // CiviCRM Initialisation
   // ---------------------------------------------------------------------------
 
+  protected function assertPhpSupport() {
+    // Need to check this before bootstrapping - once we start bootstrapping, the error messages will become ugly.
+    if ( version_compare( PHP_VERSION, CIVICRM_WP_PHP_MINIMUM ) < 0 ) {
+      echo '<p>' .
+         sprintf(
+          __( 'CiviCRM requires PHP version %s or greater. You are running PHP version %s', 'civicrm' ),
+          CIVICRM_WP_PHP_MINIMUM,
+          PHP_VERSION
+         ) .
+         '<p>';
+      exit();
+    }
+  }
 
   /**
    * Initialize CiviCRM.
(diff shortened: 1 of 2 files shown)


kcristiano:
@totten here is the backport PR

totten:
Danke schoen. `r-run` locally on a 5.16 build (alongside https://github.com/civicrm/civicrm-core/pull/15090), and it worked the same as on 5.16. :+1:

christianwach:
@totten Just a pedantic note to say that it would be nice to have a docblock for the `assertPhpSupport` method :)

Also, when using `sprintf` with more than one substitution, one should denote the placeholders as `%1$s`, `%2$s`, and so on -- this is because some languages may alter the order in which substitutions take place.

kcristiano:
@christianwach good points.   Once this is merged against master let's do a cleanup PR.  While this solves the immediate regression your changes should be made.

totten:
@christianwach - Oh, yup, very good. :) I'll post a PR on `master`.

For the sprintf bit - thanks, I'd never seen that notation before. I suppose that makes `sprintf()` a better building block for l10n cases. Also without l10n, it seems useful on general cleanliness (i.e. if someone revises the prose in the future, they don't need re-order the other bits of PHP code).

christianwach:
> I suppose that makes sprintf() a better building block for l10n cases.

@totten Absolutely. WordPress uses this notation liberally - [random example here](https://github.com/WordPress/WordPress/blob/29b072e7063f935a0af99e6ee42d6d33d6538de7/wp-admin/update-core.php#L90-L91). FWIW, the comment `/* translators: ... */` in the example also [has a format](https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#descriptions) to help translators with less technical ability understand the substitutions where necessary.

My go-to resource is [this article](http://ottopress.com/2012/internationalization-youre-probably-doing-it-wrong/) which is old but still valid. Not ashamed to say that I refer to it regularly!

HTH :)

PR #165, opened 2019-10-05
redirect to setup page on activation /dev/wordpress/issues/37

kcristiano:
Signed-off-by: Kevin Cristiano <kcristiano@kcristiano.com>

Overview
It has become expected behavior for plugins that are more complex to redirect to their install or setup screen on activation.  This will redirect on activation but not on bulk activation

Before
Users have to click on the CiviCRM notice at the top of the site or in the sidebar

After
On activation redirect users to the install screen .  

Comments
https://lab.civicrm.org/dev/wordpress/issues/37

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -310,7 +310,10 @@ class CiviCRM_For_WordPress {
 
     // Change option so this action never fires again
     update_option( 'civicrm_activation_in_progress', 'false' );
-
+    if (!isset($_GET['activate-multi'])) {
+      wp_redirect(admin_url("options-general.php?page=civicrm-install"));
+      exit;
+    }
   }
 
 
--- a/civicrm.php
+++ b/civicrm.php
@@ -310,7 +310,7 @@ class CiviCRM_For_WordPress {
 
     // Change option so this action never fires again
     update_option( 'civicrm_activation_in_progress', 'false' );
-    if (!isset($_GET['activate-multi'])) {
+    if ( ! is_multisite() && !isset($_GET['activate-multi'])) {
       wp_redirect(admin_url("options-general.php?page=civicrm-install"));
       exit;
     }


civicrm.php
```diff
@@ -307,6 +308,13 @@ public function activation() {
      * @since 5.6
      */
     do_action( 'civicrm_activation' );
+    if (get_option('civicrm_activation_in_progress', false)) {
+      delete_option('civicrm_activation_in_progress');
+      if (!isset($_GET['activate-multi'])) {
+        wp_redirect(admin_url("options-general.php?page=civicrm-install"));
+        exit;
+      }
+    }
```
christianwach: @kcristiano I think this block needs to come after:

`update_option( 'civicrm_activation_in_progress', 'false' );`

Otherwise that option will never be updated.

christianwach: Is it not possible just to add:

```php
if (!isset($_GET['activate-multi'])) {
  wp_redirect(admin_url("options-general.php?page=civicrm-install"));
  exit;
}
```

to the end of this method?

kcristiano: That does simplfy the code .  I tend to do things the long way.

kcristiano:
@christianwach Updated based on comments, rebased and squashed.

civicrm.php
christianwach: @kcristiano Just wondering if the `civicrm_setup_do_activation_redirect` option is needed at all here? The flow is already handled by the `civicrm_activation_in_progress` option (which intentionally remains in place until deactivation) so I'm wondering what `civicrm_setup_do_activation_redirect` achieves here.

kcristiano: The idea was to have it in case the redirect failed to execute.  But I have not had any issues in testing.  Perhaps we should just remove.

christianwach:
@kcristiano Looks good to me, but I'll test to be certain.

christianwach:
@kcristiano Works great in single site WordPress.

kcristiano:
@christianwach Great point.  

I think that we need to prevent the redirect from occurring on Network Activation.   

Do we want to prevent redirect even when activating on single sites?  We would not know their setup so I think we should.

Do you agree?

christianwach:
@kcristiano Network-activation could legitimately redirect to the installer on the main site (which is what it would do with this PR as it stands) and I don't think anyone would be overly surprised at the change of context. It would be preferable if the installer lived in Network Admin in this situation, but that's a bit out of scope here!

I share your concern about main site or sub-site activation when CiviCRM is not network-activated. I agree that an automatic redirect should be avoided in this scenario.

kcristiano:
I could prevent redirect on `is_multisite()`  let me look at what would need to be done to prevent just single site activation.   

Granted if it is the second site a settings file would exist and could/should prevent the installer, but taht also depends on each Mult-Domain/Multi-Site setup.

christianwach:
> I could prevent redirect on `is_multisite()`

@kcristiano Detecting network-activation might be awkward ([here's how I test this in CAU](https://github.com/christianwach/civicrm-admin-utilities/blob/c09801ad505ecfeba9ff5095133f139e70626588/civicrm-admin-utilities.php#L290-L333)) but in this plugin it's especially problematic since the `activation()` method doesn't actually run on activation, but on the next page load thereafter.

As a result, I doubt that `$_GET['activate-multi']` is ever set, so the redirect will *always* happen. This might actually be a good use of the extra option that you had in the code previously - i.e. to store the value of `$_GET['activate-multi']` in `activate()` for use in `activation()`.

> Granted if it is the second site a settings file would exist and could/should prevent the installer, but that also depends on each Mult-Domain/Multi-Site setup.

Quite so. It looks to me like the proverbial can of worms. Personally, I'd just prevent redirection in all cases when it's Multisite.

kcristiano:
@christianwach I made the change and it works as expected on both single site and multi-site.  Can you give it a test.

christianwach:
@kcristiano Did you check whether or not `$_GET['activate-multi']` ever has a value? I'm pretty sure that it needs to be stored in `activate()` then retrieved in `activation()`.

kcristiano:
I did and it does but only when activating multiple plugins.   That stops the redirect when more than 1 plugin is activated.  Or at least it did on my tests , are you finding something different?

christianwach:
@kcristiano Here is the bulk plugin activation behaviour in Multisite...

### Multisite bulk network-activation:

```
[11-Oct-2019 11:12:42 UTC] Array
(
    [method] => CiviCRM_For_WordPress::activate
    [GET] => Array
        (
            [networkwide] => 1
        )

)

[11-Oct-2019 11:12:42 UTC] Array
(
    [method] => CiviCRM_For_WordPress::activation
    [GET] => Array
        (
            [activate-multi] => true
            [plugin_status] => all
            [paged] => 1
            [s] => 
        )
)
```

### Multisite bulk single site activation.

```
[11-Oct-2019 11:19:47 UTC] Array
(
    [method] => CiviCRM_For_WordPress::activate
    [GET] => Array
        (
        )

)

[11-Oct-2019 11:19:47 UTC] Array
(
    [method] => CiviCRM_For_WordPress::activation
    [GET] => Array
        (
            [activate-multi] => true
            [plugin_status] => all
            [paged] => 1
            [s] => 
        )

    [activate-multi] => true
)
```

On which basis, I reckon your code is good to merge 👍

In case we want to revisit this in future (to refine what happens in different activation scenarios) it might be worth adding a comment to the docblock for `activate()` to the effect that network-activation can only be detected in `activate()` and not in `activation()`. Or perhaps just an `@see https://github.com/civicrm/civicrm-wordpress/pull/165` reference.

christianwach:
Just for completeness, here's the signature when CiviCRM is activated on its own:

### Single site activation:

```
[11-Oct-2019 13:54:14 UTC] Array
(
    [method] => CiviCRM_For_WordPress::activate
    [GET] => Array
        (
            [action] => activate
            [plugin] => civicrm/civicrm.php
            [plugin_status] => all
            [paged] => 1
            [s] => 
            [_wpnonce] => 78c29e3ed9
        )

)

[11-Oct-2019 13:54:14 UTC] Array
(
    [method] => CiviCRM_For_WordPress::activation
    [GET] => Array
        (
            [activate] => true
            [plugin_status] => all
            [paged] => 1
            [s] => 
        )
)
```

PR #169, opened 2019-11-14
(NFC) (dev/core#878) Simplify copyright header

totten:
https://lab.civicrm.org/dev/core/issues/878

### DIFF
--- a/assets/templates/civicrm.shortcode.php
+++ b/assets/templates/civicrm.shortcode.php
@@ -2,34 +2,18 @@
 
 /*
  +--------------------------------------------------------------------+
- | CiviCRM version 5                                                  |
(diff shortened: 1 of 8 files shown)


kcristiano:
Should we move the above details into the standard Plugin header?  That will be more the WP style.   @christianwach Thoughts?   Is that better or should be be consistent with all other core files.

christianwach:
@kcristiano It's already declared where needed:

https://github.com/civicrm/civicrm-wordpress/search?q=AGPL3&unscoped_q=AGPL3

I'm happy for the new standard headers to be applied to each file.

totten:
That sounds likes a 'merge on pass' to me. :)

PR #171, opened 2019-11-25
Set CMS Root in "civicrm_paths" more reliably

christianwach:
Overview
Builds on https://lab.civicrm.org/dev/core/issues/1412 to set `$civicrm_paths['cms.root']['path']` more reliably during the following requests:

* WP-CLI
* "Heartbeat" AJAX
* WordPress pseudo-cron

Before
`CRM_Utils_System_WordPress::cmsRootPath()` would be called without `$civicrm_paths['cms.root']['path']` being set for the above requests, causing the unreliable directory-traversal code to be run.

After
`CRM_Utils_System_WordPress::cmsRootPath()` runs with `$civicrm_paths['cms.root']['path']` being correctly set.

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -364,6 +364,15 @@ class CiviCRM_For_WordPress {
       session_start();
     }
 
+    /*
+     * AJAX calls do not set the 'cms.root' item, so make sure it is set here so
+     * the CiviCRM doesn't fall back on flaky directory traversal code.
+     */
+    global $civicrm_paths;
+    if (empty($civicrm_paths['cms.root']['path'])) {
+      $civicrm_paths['cms.root']['path'] = untrailingslashit(ABSPATH);
+    }
+
     // Get classes and instantiate
     $this->include_files();
 
--- a/wp-cli/civicrm.php
+++ b/wp-cli/civicrm.php
@@ -165,13 +165,6 @@ if ( ! defined( 'CIVICRM_WPCLI_LOADED' ) ) {
         return WP_CLI::error( "Unrecognized command - '$command'" );
       }
 
-      # if --path is set, save for later use by Civi
-      global $civicrm_paths;
-      $wp_cli_config = WP_CLI::get_config();
-      if (!empty($wp_cli_config['path'])) {
-        $civicrm_paths['cms.root']['path'] = $wp_cli_config['path'];
-      }
-
       # run command
       return $this->{$command_router[ $command ]}();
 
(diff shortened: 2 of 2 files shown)


kcristiano:
* ([`r-explain`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-explain))Not needed
* ([`r-test`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-test)) Pass
* ([`r-code`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-code)) Pass
* ([`r-doc`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-doc)) Pass
* ([`r-run`](https://docs.civicrm.org/dev/en/latest/standards/review/#r-run)) Pass: 
- Standard installl  with PR Applied.
  - pass on mailings, searches, contrib pages, wp-cli, and cv 
- WP in it's own directory
  - pass on mailings, searches, contrib pages, wp-cli, and cv
- WP in it's own directory and wp-content moved
  - pass on mailings, searches, contrib pages, wp-cli, and cv


This looks good to merge.

@seamuslee001 any further thoughts?

kcristiano:
As far as testing:

- "Heartbeat" AJAX
- WordPress pseudo-cron

@christianwach Are there any specific tests that you can think of?  

wp cron is running properly in my test scenarios, heartbeat is not tossing errors.   Is there an example where CiviCRM uses admin-ajax?  I don't see it, am I missing the obvious?

christianwach:
> As far as testing:
> 
> * "Heartbeat" AJAX
> * WordPress pseudo-cron
>
> @christianwach Are there any specific tests that you can think of?

### "Heartbeat" AJAX

I can't think of anything specific right now, but that's not to say that other plugins might not hook into Heartbeat operations - for example doing something when a draft post is saved. Given we have been thinking about CiviCRM Entities syncing with WordPress CPTs, this might prove an important fix in future.

### WordPress pseudo-cron

Again, struggling to think of anything that's guaranteed to be affected. One of the options in CiviCRM WP Member Sync allows it to hook into WordPress pseudo-cron and [trigger processes that rely on proper CiviCRM configuration](https://github.com/christianwach/civicrm-wp-member-sync/blob/a48ba667de03c7cab492c529225c2e7cb65dbd1d/includes/civi-wp-ms-schedule.php#L129-L139). That it works as it stands would seem to be fortuitous rather than by design.

I think the tests is simply that `$civicrm_paths['cms.root']['path']` is correctly set when triggered via these routes. I have confirmed that it is. Throw some error logging in at the very top of `CRM_Utils_System_WordPress::cmsRootPath()` if you want to confirm this:

```php
    $e = new Exception;
    $trace = $e->getTraceAsString();
    error_log( print_r( array(
      'method' => __METHOD__,
      'civicrm_paths' => $civicrm_paths,
      'backtrace' => $trace,
    ), true ) );
```

The backtrace will show you the origin of the request and as long as `$civicrm_paths` is populated, then all's well.

kcristiano:
Thanks @christianwach   Based on the testing, I'll merge

PR #174, opened 2019-12-03
Fix basepage title for Yoast SEO when theme declares "title-tag" support

christianwach:
As reported [on Mattermost](https://chat.civicrm.org/civicrm/pl/tjbhpekwmpdjmp7733irnwij5y) by Marcus J Wilson.

Overview
Fixes the basepage title when the Yoast SEO plugin is active and the theme declares "title-tag" support via `add_theme_support( 'title-tag' )`.

Before
The Yoast SEO plugin short-circuits the code in `wp_get_document_title()` by filtering via `pre_get_document_title`. This means that our code to apply the CiviCRM basepage title via the `document_title_parts` filter is never reached and the basepage title does not reflect the CiviCRM content.

After
The Yoast SEO plugin no longer parses the title and our code runs as expected.


kcristiano:
I did an `r-run` and this works as described

PR #175, opened 2019-12-16
Update to use the namespaced version of the PHPUnit test classes as t…

seamuslee001:
…his will work for phpunit6 going forward as well as phpunit5

Overview
This updates the class names to be their namespaced equivalents 

ping @kcristiano @totten


kcristiano:
@seamuslee001 is there a related core PR we need to wait for?  This looks good to merge to me

seamuslee001:
@kcristiano no

PR #178, opened 2020-02-12
Make sure Rewrite Rules are flushed after creating the Base Page

christianwach:
Overview
----------------------------------------
Rewrite Rules have been defined and flushed prior to the creation of the Base Page resulting in 404s for CiviCRM pseudo-pages.

Before
----------------------------------------
404s on the Base Page.

After
----------------------------------------
Base Page content renders correctly.

PR #179, opened 2020-02-12
Duplicate of "Make sure Rewrite Rules are flushed after creating the Base Page"

christianwach:
Duplicate of #178


kcristiano:
test fail unrelated.

r-run:  PASS

Without this PR you need to load CiviCRM in the admin and refresh the rewrite rules before CiviCRM pages will load properly with Cleaner URLs.

A larger scope would be to determine away to not have the initial load be necessary fpr automated installs.

PR #182, opened 2020-02-14
[NFC] Housekeeping (5.23 branch)

christianwach:
Overview

Duplicate of #181 against the 5.23 branch.


kcristiano:
@christianwach do we want this in 5.23 or master?   If 5.23 we should merge this and then do a PR to master.

christianwach:
@kcristiano It should be in both, but I'm not fussy about how that happens :-)

languages/civicrm.pot
```diff
@@ -1,192 +1,115 @@
-# Copyright (C) 2014 CiviCRM
+# Copyright (C) 2020 CiviCRM
 # This file is distributed under the same license as the CiviCRM package.
 msgid ""
 msgstr ""
-"Project-Id-Version: CiviCRM 4.6\n"
-"Report-Msgid-Bugs-To: http://wordpress.org/tag/civicrm\n"
-"POT-Creation-Date: 2014-11-11 09:48:56+00:00\n"
+"Project-Id-Version: CiviCRM 4.7\n"
```
JoeMurray: I'm not that familiar with what's required here. Just checking you are putting in 4.7 on the basis that the translations are good for 4.7+, rather than 5.x or 5.23 which might otherwise make more sense.

christianwach: @JoeMurray This file is generated by WP-CLI or the WordPress `makepot` bundle (the latter in this case) and the stated version is read from either [the entry in `readme.txt`](https://github.com/civicrm/civicrm-wordpress/blob/master/readme.txt#L6) or [the declared plugin version](https://github.com/civicrm/civicrm-wordpress/blob/master/civicrm.php#L5). 

It's not actually relevant to the plugin at this stage, since it's not hosted in the WordPress Plugin Directory. Refreshing this file just allows plugins like Loco Translate to pick up all (non CiviCRM Core) translatable strings in this plugin.

kcristiano:
test this please

languages/civicrm.pot
```diff
@@ -1,192 +1,115 @@
...
+"Report-Msgid-Bugs-To: https://wordpress.org/support/plugin/civicrm\n"
```
kcristiano: @christianwach this link is not active.  Is this in anticipation of the plugin repo accepting the submission?

christianwach: @kcristiano Please see [my reply to Joe](https://github.com/civicrm/civicrm-wordpress/pull/182#discussion_r379571956) which answers this.

kcristiano: Got it.

PR #188, opened 2020-03-24
Set "cms.root" URL in addition to Path

christianwach:
Overview
The code in `Civi::Paths::getVariable()` does not ensure all attributes of a named variable are set, so partially setting the attributes of `cms.root` sometimes causes "Cannot resolve path using 'cms.root.url'" exceptions to be thrown. This is relatively obscure because it requires a particular WordPress Multisite setup to occur, but the effect is repeatable.

Before
"Cannot resolve path using 'cms.root.url'" exceptions are thrown. "Settings - Resource URLs" screen renders incorrectly in WordPress Multisite on a Subsite:

Screen Shot 2020-03-24 at 11 01 40

After
"Cannot resolve path using 'cms.root.url'" exceptions are not thrown. "Settings - Resource URLs" screen renders correctly in WordPress Multisite on a Subsite.

Technical Details
This PR is necessary because of an incomplete check for the existence of all attributes [in Paths.php](https://github.com/civicrm/civicrm-core/blob/c879543be97e2d851430db3f17afaf9f4a1d00fa/Civi/Core/Paths.php#L154). Having said that, it's still better to assign both attributes early than have `Civi::Paths::getVariable()` recalculate them later.

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -368,6 +368,9 @@ class CiviCRM_For_WordPress {
     if (empty($civicrm_paths['cms.root']['path'])) {
       $civicrm_paths['cms.root']['path'] = untrailingslashit(ABSPATH);
     }
+    if (empty($civicrm_paths['cms.root']['url'])) {
+      $civicrm_paths['cms.root']['url'] = home_url();
+    }
 
     // Get classes and instantiate
     $this->include_files();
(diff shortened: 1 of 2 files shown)


kcristiano:
code is sensible.   Should have been there from the start.

I did an `r-run` on master and works as expected.

PR #192, opened 2020-05-14
wp-rest - Adjust extern URL. Be more defensive about 'query' part.

totten:
Overview
--------

`hook_civicrm_alterExternUrl` allows for modifications to certain callback URLs (e.g.  CiviMail open/click-through URLs). As a hook, there may be multiple parties which weigh-in.

This fixes an interaction that arises in the following conditions:

* Use Civi+WordPress
* Set `CIVICRM_WP_REST_REPLACE_MAILING_TRACKING` to `TRUE`
* Use https://github.com/civicrm/civicrm-core/pull/17312
* Send a mailing (or test email) with a tracked link

Before
------

The tracked link is converted to something like

```
http://wpmaster.bknix:8001/wp-json/civicrm/v3/url?page=CiviCRM&q=civicrm/mailing/url
  &u=5&qid=67
```

On the plus side, this includes query parameters `u=5&qid=67` - which are important inputs for any `extern/url`-style end-point.

On the downside, the URL is mixing up artifacts which identify different end-points, ie

* `wp-json/civicrm/v3/url` suggests the `wp-rest` end-point
* `?page=CiviCRM&q=civicrm/mailing/url` suggests the #17312 end-point

After
-----

The tracked link appears as:

```
http://wpmaster.bknix:8001/wp-json/civicrm/v3/url?u=5&qid=68
```

Notice that this unambiguously uses the `wp-rest` end-point.  It includes the important `u=5&qid=67`.

Technical Details
-----------------

The contract for `hook_civicrm_alterExternUrl` allows multiple parties to weigh-in on the URL. In this case, we have 3 parties which can generate the URL:

* One function suggests the original standalone scripts
* Another function suggests the #17312 routes
* Another function suggests the `wp-rest` routes

This is prioritized such that `wp-rest` will override the URLs suggested by #17312.

Aside: When using `alterExternUrl` to change the end-point, it seems perhaps most robust to build `$url` from scratch -- ie don't read anything from `$url`, read it from the request-values in `$path` and `$query`. I've suggested some updates to the examples: https://github.com/civicrm/civicrm-dev-docs/pull/807

### DIFF
--- a/wp-rest/Civi/Mailing-Hooks.php
+++ b/wp-rest/Civi/Mailing-Hooks.php
@@ -80,12 +80,14 @@ class Mailing_Hooks {
 		if ( $path == 'extern/url' ) {
 			$url = $url
 				->withHost( $this->parsed_rest_url['host'] )
+				->withQuery( $query )
 				->withPath( "{$this->parsed_rest_url['path']}civicrm/v3/url" );
 		}
 
 		if ( $path == 'extern/open' ) {
 			$url = $url
 				->withHost( $this->parsed_rest_url['host'] )
+				->withQuery( $query )
 				->withPath( "{$this->parsed_rest_url['path']}civicrm/v3/open" );
 		}
 


totten:
jenkins, test this please

kcristiano:
@totten Thanks I'll test.

I also want @mecachisenros to chime in on this

eileenmcnaughton:
Merged per @kcristiano comment regarding review on https://github.com/civicrm/civicrm-core/pull/17312

PR #196, opened 2020-05-20
dev/wordpress#53 Replace tracking URLs only if mail content is replaceable

mecachisenros:
Overview
See [dev/wordpress#53](https://lab.civicrm.org/dev/wordpress/-/issues/53)

Checks mailing content is replaceable before attempting to replace tracking URLs.

Before
The mailing parameters `text` and `html` could be `NULL` or not defined, throwing an error when attempting to replace the tracking URLs. 

After
If the mailing content is not replaceable there's no attempt to replace it.

### DIFF
--- a/wp-rest/Civi/Mailing-Hooks.php
+++ b/wp-rest/Civi/Mailing-Hooks.php
@@ -106,13 +106,17 @@ class Mailing_Hooks {
 
 		if ( in_array( $context, [ 'civimail', 'flexmailer' ] ) ) {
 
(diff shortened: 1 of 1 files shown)


kcristiano:
@andyburnsco  Can you give this one a look as well?

andyburnsco:
Will test tonight

andyburnsco:
My mailing was successful, it sent to all 1615 recipients. Thanks @mecachisenros!

PR #194, opened 2020-05-18
Remove "page=CiviCRM" query string from WordPress front-end (5.26)

christianwach:
Overview
----------------------------------------
Additional PR against 5.26 branch. Same fix as #193. Requires [this PR on CiviCRM Core](https://github.com/civicrm/civicrm-core/pull/17352). Solves [this issue on Lab](https://lab.civicrm.org/dev/wordpress/-/issues/49).

PR #199, opened 2020-05-30
dev/wordpress#49 Prevent 301 redirects if 'page=CiviCRM'

mecachisenros:
Overview
Prevents old URLs like `example.org/civicrm/?page=CiviCRM&q=what/ever/path&reset=1`
being redirected to `example.org/civicrm/?civiwp=CiviCRM&q=what/ever/path&reset=1`.

Before
Old URLs like `example.org/civicrm/?page=CiviCRM&q=what/ever/path&reset=1` would be 301 redirected.

After
No redirects.

Technical Details
Filters the request right after WP's parsed it and replaces the 'page' query variable with 'civiwp' if applicable.

Comments
I think this should be an addition to @christianwach work on #194 because one could prevent WP from parsing the request (via hooks) which will not trigger this.

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -578,6 +578,9 @@ class CiviCRM_For_WordPress {
     // Go no further if CiviCRM not installed yet
     if ( ! CIVICRM_INSTALLED ) return;
 
+    // Attempt to replace 'page' query arg with 'civiwp'.
+    add_filter( 'request', array( $this, 'maybe_replace_page_query_var' ) );
+
     // Add our query vars.
     add_filter( 'query_vars', array( $this, 'query_vars' ) );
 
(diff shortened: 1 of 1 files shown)


christianwach:
@mecachisenros Thanks for starting on this 👍 I think you can remove [this code](https://github.com/civicrm/civicrm-wordpress/blob/396d83623b17751027a5b9d76c60d097fb472943/civicrm.php#L619-L625) which should be unnecessary with you additions.

mecachisenros:
@christianwach I think that code should remain as a fallback, the reason being one could [bypass](https://developer.wordpress.org/reference/hooks/do_parse_request/) WP from parsing the query which would prevent my code from running, and that would _break_ the page.

christianwach:
@mecachisenros Gotcha - fair point.

kcristiano:
Did `r-run` on 5.26 site and all worked as expected.

I also did an `r-run` with a site upgraded from 5.25 to 5.26 and I could not 'break it.

I have created recurring payments to track if this will allow existing URLs in the old page format to redirect with payload.

Either way, this is far better merged.

PR #201, opened 2020-06-06
Declare Minimum WP and PHP versions required for installation

kcristiano:
Overview
CiviCRM will fail to install on WP Versions less than 4.7  due to https://github.com/civicrm/civicrm-core/commit/742eb5c659cce650a3d3739ba341b065266db714

Based on https://lab.civicrm.org/dev/wordpress/-/issues/51  we want to update minimum WP version to 4.9

Before
CiviCRM install will fail on versions older than 4.7

After
CiviCRM cannot be activated unless WP meets the minimum requirement.  Added a php minimum as well

Technical Details

Comments
Once accepted we must update the docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,11 +9,14 @@ This is the development repository for the *CiviCRM* plugin for *WordPress*. Wha
 
 If you want to contribute to the development of this plugin, please bear the following in mind:
 
-* Bug fixes should go in the branch 4.5 branch (stable)* Structural changes should go under master (trunk, i.e. 4.6).
+* Bug Fixes and structural changes should go in the master branch.
+* Regression fixes should go in the current version branch.
 
 ----
-### About CiviCRM ###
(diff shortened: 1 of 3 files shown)


MikeyMJCO:
This has already been documented by @colemanw :-)

MikeyMJCO:
https://docs.civicrm.org/sysadmin/en/latest/requirements/

kcristiano:
@seamuslee001 I've added a WP check for php version.   This will prevent activation on php < 7.1   Currently the plugin will activate and errors will occur in the installer.

MikeyMJCO:
A quick buildkit test shows that this works.

kcristiano:
One further note - I've targeted 5.26 as CiviCRM will fail on versions older than 4.7 currently which is an unintended regression.

kcristiano:
Just to note here:   Certain versions are showing 4.7 - these get replaced with the current version during the build process so I have left them as is.

PR #203, opened 2020-06-09
dev/core#1681 Block sites from being able to upgrade CiviCRM if runni…

seamuslee001:
…ng MySQL versions below 5.6.5

ping @totten @kcristiano

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -89,6 +89,22 @@ if ( !defined( 'CIVICRM_WP_PHP_MINIMUM' ) ) {
   define( 'CIVICRM_WP_PHP_MINIMUM', '7.1.0' );
 }
 
+if ( !defined( 'CIVICRM_WP_MYSQL_MINIMUM' ) ) {
+  /**
+   * Minimum required MySQL
+   *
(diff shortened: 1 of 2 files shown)


seamuslee001:
Jenkins re test this please

kcristiano:
I've done an `r-run` and this PR does work as intended.   I am unclear if this will prevent MariaDB 10.0 (which IIRC is mysql 5.5 compatible) from loading properly.  I think any maria versions above 10.1 will be fine, but I am unclear about 10.0.  

Also see - https://chat.civicrm.org/civicrm/pl/idcorgxngiygmqbzqmk7pc8e1r  for a discussion on whether we want the check on all page loads.  My inclination is we are better off testing on install upgrade.  The only riskj I see is if a site is moved from a later version of MySQL to an earlier one - but that should cuase pain unrelated to CiviCRM.

seamuslee001:
@totten ^

seamuslee001:
Closing this as per comment https://github.com/civicrm/civicrm-core/pull/17558#issuecomment-646932490

PR #204, opened 2020-06-11
Fix discovery of page templates in theme subdirectories

christianwach:
Overview
Currently, the CiviCRM base page fails to apply the selected Page Template if that template does not reside in the root directory of the theme. The fallback `page.php` template is used instead.

Fixes [this issue on Lab](https://lab.civicrm.org/dev/wordpress/-/issues/58).


kcristiano:
Performed `r-run` and this handles well the case detailed in this issue on Lab](https://lab.civicrm.org/dev/wordpress/-/issues/58) without affecting other configurations

PR #214, opened 2020-07-22
Slow down the frequency of WordPress "heartbeat" calls in CiviCRM admin

christianwach:
Overview
Heartbeat is important to WordPress for a number of tasks - e.g. checking continued authentication whilst on a page and implementing edit locks on post/page edit screens - but it is less important on CiviCRM admin pages and consumes server resources. Reducing the frequency of calls minimises the impact on servers and can make CiviCRM admin pages more responsive.

[Lab issue here](https://lab.civicrm.org/dev/wordpress/-/issues/67)

Before
Heartbeat calls to the server happen every 15 seconds.

After
Heartbeat calls to the server happen every 120 seconds.

Comments
This can be tested by logging access to the method introduced in this PR. Sample code:

```php
    error_log( print_r( array(
      'method' => __METHOD__,
      'existing-heartbeat-settings' => $settings,
    ), true ) );
```

If you're wondering why the method needs to parse the requested URL, it's because the `heartbeat_settings` filter is called before admin utility methods such as `get_current_screen()` are available.

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -387,6 +387,9 @@ class CiviCRM_For_WordPress {
     // Register all hooks on init
     add_action( 'init', array( $this, 'register_hooks' ) );
 
+    // Filter Heartbeat on CiviCRM admin pages as late as is practical.
+    add_filter( 'heartbeat_settings', array( $this, 'heartbeat' ), 1000, 1 );
+
     /**
      * Broadcast that this plugin is now loaded.
      *
@@ -440,6 +443,56 @@ class CiviCRM_For_WordPress {
     if (empty($session_id) && !$wp_cron && !$wp_cli && !$php_cli) {
       session_start();
     }
+
+  }
+
+
+  /**
+   * Slow down the frequency of WordPress heartbeat calls.
+   *
+   * Heartbeat is important to WordPress for a number of tasks - e.g. checking
+   * continued authentication whilst on a page - but it does consume server
+   * resources. Reducing the frequency of calls minimises the impact on servers
+   * and can make CiviCRM more responsive.
+   *
+   * @since 5.29
+   *
+   * @param array $settings The existing heartbeat settings.
+   * @return array $settings The modified heartbeat settings.
+   */
+  public function heartbeat( $settings ) {
+
+    // Access script identifier.
+    global $pagenow;
+
+    // Bail if not admin.
+    if (!is_admin()) {
+      return $settings;
+    }
+
+    // Process the requested URL.
+    $requested_url  = filter_input(INPUT_SERVER, 'REQUEST_URI', FILTER_SANITIZE_URL);
+    if ( $requested_url ) {
+      $current_url = wp_unslash($requested_url);
+    } else {
+      $current_url = admin_url();
+    }
+    $current_screen = wp_parse_url($current_url);
+
+    // Bail if this is not CiviCRM admin.
+    if ($pagenow != 'admin.php' || false === strpos($current_screen['query'], 'page=CiviCRM')) {
+      return $settings;
+    }
+
+    // Defer to any previously set value, but only if it's greater than ours.
+    if (!empty($settings['interval']) && intval($settings['interval']) > 120) {
+      return $settings;
+    }
+
+    // Slow down heartbeat.
+    $settings['interval'] = 120;
+
+    return $settings;
 
   }
 


kcristiano:
@christianwach My only concern is what if the site already has heartbeat modified with the same filter?   May shared hosts have done this and https://wordpress.org/plugins/heartbeat-control/  as well as many caching plugins alter it's value as well.

I did an `r-run` with the heart beat control plugin and that seems to be the filter that is being used along with this change.

Can you think of any trouble that will cause?

kcristiano:
@christianwach just looking at how I setup heartbeat, I tend to disable everywhere but on `edit-post`  any reason not to disable heartbeat for CiviCRM admin pages?

christianwach:
@kcristiano Hmm, yes, there will be some level conflict when using Heartbeat Control as this PR currently stands. Heartbeat Control allows modification of `heartbeat` in a couple of ways:

* It can _disable_ it on one or more screens - this is done badly in Heartbeat Control IMO because the WordPress `wp-auth` script then complains about `heartbeat` being a missing dependency. That's why this PR _only_ touches the `interval` setting - and, if the script has been deregistered, it will have no effect.
* It sets the `interval` on one or more screens - here's where this PR may conflict. The question here is whether the setting here ought to _override_ or _defer_ to the Heartbeat Control setting. I'm open to suggestions on that and can update this PR accordingly.

christianwach:
> I tend to disable everywhere but on edit-post any reason not to disable heartbeat for CiviCRM admin pages?

@kcristiano WordPress auth checks fail when `heartbeat` is disabled. This is a potential (though admittedly low) security risk, so I wouldn't do it myself.

christianwach:
> many caching plugins alter it's value as well

@kcristiano I didn't know this. Can you point me to some and I'll take a look?

kcristiano:
@christianwach wp rocket for sure, I think W3 Total cache may have added.   Rocket is the one that tripped me up.   

My real concern is hosts.  WP Engine, Kinsta, and Siteground recommend disabling or restricting heartbeat.  I think (but can't confirm) that WPE disables on all but edit screen silently.

christianwach:
> WP Engine, Kinsta, and Siteground recommend disabling or restricting heartbeat.

@kcristiano That's fine though - if `heartbeat` is disabled, the code in this PR won't re-enable it. This PR only sets the `interval` to the maximum that `heartbeat` will accept when it is enqueued, so it is effectively restricting as much as is possible on CiviCRM's admin screens.

The open question is really the priority given to the `interval` settings _when they are set elsewhere_ and whether they should take priority over what's set here. I can adjust that very easily here if I can see their hook priorities.

kcristiano:
> The open question is really the priority given to the interval settings when they are set elsewhere and whether they should take priority over what's set here. I can adjust that very easily here if I can see their hook priorities.

I only have access to wp-rocket and heartbeat and both are set to '99'

I've run into this before, not often, usually when heartbeat was completely turned off on shared hosts.

> The question here is whether the setting here ought to override or defer to the Heartbeat Control setting. I'm open to suggestions on that and can update this PR accordingly.

My inclination is to defer.   I think your limit if 120 makes sense if nothing is already set.

christianwach:
> I only have access to wp-rocket and heartbeat and both are set to '99'

Thanks @kcristiano - what do they set the heartbeat `interval` to?

kcristiano:
The default interval is 15 seconds in the plugin and then you can set the levels.    

As I've run into this on a few shared hosts, the one's I recall that hard articles are:  

- https://www.inmotionhosting.com/support/edu/wordpress/install-and-configure-heartbeat-control-for-wordpress/  300 seconds
- https://kinsta.com/blog/admin-ajax/ - 60 seconds
- https://www.siteground.com/tutorials/wordpress/limit-heartbeat/   60 seconds or more

These are the 3 I've seen more than others.

christianwach:
@kcristiano Thanks for the figures. I've modified this PR so that an existing `interval` is respected as long as its greater than the one being set here.

> 300 seconds

The [maximum value for the interval is 120 seconds](https://github.com/WordPress/WordPress/blob/703d5bdc8deb17781e9c6d8f0dd7e2c6b6353885/wp-includes/js/heartbeat.js#L604), so I'm not sure what they think they're achieving :-)

christianwach:
>> The question here is whether the setting here ought to override or defer to the Heartbeat Control setting. I'm open to suggestions on that and can update this PR accordingly.
>
> My inclination is to defer. I think your limit if 120 makes sense if nothing is already set.

Just to clarify, I don't see any need for a _shorter_ `interval` on CiviCRM admin pages, so deferring to Heartbeat Control or any other setting when it's shorter than `120` doesn't make sense to me.

kcristiano:
> any other setting when it's shorter than 120 doesn't make sense to me.

Looking at https://github.com/WordPress/WordPress/blob/703d5bdc8deb17781e9c6d8f0dd7e2c6b6353885/wp-includes/js/heartbeat.js#L125   I expect that when this was written Jeff has the setting as a way to turn it up to 11.  Makes you feel good, but nothing happens.

As we only affect our pages, I am OK with our setting trumping others.

christianwach:
Thanks @kcristiano

PR #215, opened 2020-08-11
dev/wordpress#37 - Switch unambiguously to new installer UI

totten:
Overview
-------------

See https://lab.civicrm.org/dev/wordpress/-/issues/37

Before
------

To activate the new installer UI on WordPress, you need to either use the `civicrm-X.Y.Z-wporg.zip`
or manually create the file `.use-civicrm-setup`.

Otherwise, it uses the old installer.

Screenshot from 2020-08-11 16-13-51

After
-----

The old installer is not available on WordPress. It only uses the new installer.

Screenshot from 2020-08-11 16-13-31

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -1304,7 +1304,7 @@ class CiviCRM_For_WordPress {
     );
     foreach ($setupPaths as $setupPath) {
       $loader = implode(DIRECTORY_SEPARATOR, [$civicrmCore, $setupPath, 'civicrm-setup-autoload.php']);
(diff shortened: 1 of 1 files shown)


seamuslee001:
@kcristiano do you need any help on the testing of this one?

kcristiano:
@seamuslee001 can always use help :)   I'll make sure I hit it soon.   I'd like it in next RC

kcristiano:
I ran another test on 5.30 alpha and it looks good to me.

Is there  any reason not to merge?   I feel we have buy in and I prefer this over the standard installer, but would like some feedback.

seamuslee001:
I don't see a reason not to

PR #217, opened 2020-09-10
Add permission_callback to register_rest_route calls

bastienho:
Overview
----------------------------------------
Since release of WordPress 5.5, `permission_callback` is a mandatory parameter for `register_rest_route`

Before
----------------------------------------
A lot of PHP Warnings in logs

After
----------------------------------------
No warnings

Technical Details
----------------------------------------
This is described in https://make.wordpress.org/core/2020/07/22/rest-api-changes-in-wordpress-5-5/

PR #218, opened 2020-09-10
Add permission_callback to register_rest_route calls

bastienho:
Overview
Since release of WordPress 5.5, `permission_callback` is a mandatory parameter for `register_rest_route`

Clean PR for #217 

Before
A lot of PHP Warnings in logs

After
No warnings

Technical Details
This is described in https://make.wordpress.org/core/2020/07/22/rest-api-changes-in-wordpress-5-5/


christianwach:
@bastienho Thanks.
@kcristiano This looks good to me.
@mecachisenros Can you verify?

kcristiano:
@christianwach have you done an `r-run`?  I do think we need to target 5.30 for this, I can make that change after it has been tested and verified.

kcristiano:
I did do an `r-run` and it works as expected.   @mecachisenros validated this in a call with me.

PR #219, opened 2020-09-10
5.30

kcristiano:
Merge 5.30 branch to master


mattwire:
@kcristiano Just a quick question to confirm my understanding. The references here to specific payment processors (paypal/authorize) are they just for the old IPN URLs?

kcristiano:
@mattwire yes, this was to mirror the functionality of [extern](https://github.com/civicrm/civicrm-core/tree/master/extern) so we can use WP REST routing instead.

PR #220, opened 2020-10-12
Protect against undefined index query in heartbeat callback

christianwach:
Overview
Trivial fix for [this issue on Lab](https://lab.civicrm.org/dev/wordpress/-/issues/78).

Before
Array entry not checked before being queried.

After
Array entry checked before being queried.

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -479,6 +479,11 @@ class CiviCRM_For_WordPress {
     }
     $current_screen = wp_parse_url($current_url);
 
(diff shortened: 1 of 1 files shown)


kcristiano:
I was unable to reproduce the error on upgrade as well.   But the code looks good and makes sense.

I did an `r-run` on an upgrade from 5.30 to 5.32 (master) after building with the PR.  Works as expected.

PR #221, opened 2020-11-17
Ensure "civicrm_initialized" action only fires once

christianwach:
Overview
At present, the `civicrm_initialized` action fires every time `civi_wp()->initialize()` is called. This should only happen once.

Before
The `civicrm_initialized` action fires every time `civi_wp()->initialize()` is called.

After
The `civicrm_initialized` action fires only when `civi_wp()->initialize()` actually initialises CiviCRM.

Technical Details
Code that listens for the `civicrm_initialized` action receives callbacks every time `civi_wp()->initialize()` is called. This appears to have been an error on my part when refactoring for CiviCRM 4.6. This PR implements the intended behaviour.

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -1183,14 +1183,14 @@ class CiviCRM_For_WordPress {
 
       }
 
-    }
+      /**
+       * Broadcast that CiviCRM is now initialized.
+       *
+       * @since 4.4
+       */
+      do_action( 'civicrm_initialized' );
 
-    /**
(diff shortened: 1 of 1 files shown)


kcristiano:
I have done an `r-run`  on master with this PR applied.

Install is fine, CiviCRM itself is good.  The usual set of plugins I use run as expected.   

The change is an improvement.   If there are any plugins/extensions that are affected we'll flush them out during the RC period.

PR #223, opened 2020-11-25
[NFC] Make all files conform to CiviCRM code styling

christianwach:
Overview
Makes all files conform to CiviCRM code styling.

No functional changes.

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -1,18 +1,17 @@
 <?php
-/*
-Plugin Name: CiviCRM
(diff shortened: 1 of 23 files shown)


christianwach:
@seamuslee001 The failures are for `eval` statements in the `WP-CLI` file. These seem necessary, as far as I can tell. Is there a way to exclude those from the checks?

kcristiano:
NFC, civilint reports all OK.

Will deal with `eval` in a separate issue and plan to remove, it's been there for years.

PR #216, opened 2020-09-04
Fix wp-cli upgrade command failing without legacy settings file

gpoole:
Overview
----------------------------------------
Fixes the wp-cli `wp civicrm upgrade` command failing with the error `Unable to locate settings file at wp-content/plugins/civicrm/civicrm.settings.phpor at wp-content/uploads/civicrm/civicrm.settings.php` if only one of those two files exists.

Before
----------------------------------------

The upgrade command currently fails on my install with a `Unable to locate settings file at` error about not being able to find `civicrm.settings.php` file in one of two locations:
- `$wp_root/wp-content/plugins/civicrm/civicrm.settings.php`
- `$wp_root/wp-content/uploads/civicrm/civicrm.settings.php`

The check assume that _both_ files exist, which wasn't true for my install (I only have it in the non-legacy location).

After
----------------------------------------

The legacy settings are loaded only if the new settings file doesn't exist. If both exist, only the new settings file is used. If the legacy settings file doesn't exist, there is no error.

PR #224, opened 2020-11-25
Allow wp-cli upgrade command to proceed when there is only a single settings file

christianwach:
Overview
Fixes the immediate issue raised in #216 so that upgrades proceed.

Before
The upgrade command fails on with "Unable to locate settings file at..." unless there is a `civicrm.settings.php` file in both the legacy and current locations.

After
The upgrade command proceeds if a `civicrm.settings.php` file is found in either the legacy or the current locations.

Comments
This PR does not address what follows the settings file check but then nor does #216 either really. The consensus seems to be that when pre CiviCRM 4.7 installs are upgraded, their `civicrm.settings.php` file in the legacy location should be moved to the current `wp_uploads_dir()` location.

The [analysis by @kcristiano on #216] should be followed when fixing this in a fresh PR.


christianwach:
Jenkins failure is due to the `eval` statements - these will be resolved in the follow up PR that fixes the upgrade procedure.

seamuslee001:
Jenkins re test this please

kcristiano:
jenkins re-test this please

kcristiano:
Upgrade still fails with this code, but we need this to be able to fix the upgrader

PR #226, opened 2020-11-26
wordpress#63 Add action parameter to PCP shortcode

mattwire:
Overview
Add action parameter to PCP shortcode per https://lab.civicrm.org/dev/wordpress/-/issues/63

The parameters chosen are "info" and "transact" because I think this makes more sense than using "register" as used by event.

Requires https://github.com/civicrm/civicrm-core/pull/19058

Before
No action parameter for PCP shortcodes.

After
Action parameter for PCP shortcodes.

Technical Details

Comments
@kcristiano @christianwach

### DIFF
--- a/includes/civicrm.shortcodes.php
+++ b/includes/civicrm.shortcodes.php
@@ -704,7 +704,18 @@ class CiviCRM_For_WordPress_Shortcodes {
         if ($mode == 'preview' || $mode == 'test') {
           $args['action'] = 'preview';
         }
-        $args['q'] = 'civicrm/pcp/info';
+
+        switch ($action) {
+          case 'transact':
+            $args['q'] = 'civicrm/contribute/transact';
+            $args['pcpId'] = $args['id'];
+            break;
+
+          case 'info':
+          default:
+            $args['q'] = 'civicrm/pcp/info';
+            break;
+        }
         break;
 
       case 'event':


includes/civicrm.shortcodes.php
```diff
@@ -704,7 +704,20 @@ public function preprocess_atts($atts) {
...
+            $args['q'] = 'civicrm/contribute/transact';
+            $args['pcpId'] = $args['id'];
+            break;
+
+          case 'info':
+            $args['q'] = 'civicrm/pcp/info';
+            break;
+
+          default:
+            return FALSE;
```
christianwach, line 719: @mattwire Shouldn't the default be the same as `case 'info':` to preserve existing behaviour?

mattwire: @christianwach Yes agreed. I've left the case 'info': in and a fall-through to default. I know that's not needed but it seems clearer to someone reading the code.

christianwach: @mattwire Agreed, looks good now.

kcristiano:
This needs to be merged with https://github.com/civicrm/civicrm-core/pull/19058

@christianwach has reviewed and I have done an `r-run` and it passes.  This is merge ready.

kcristiano:
Reference :  https://lab.civicrm.org/dev/core/-/issues/2278

PR #227, opened 2020-11-26
Fix assignment of base page template

christianwach:
Overview
I somehow managed to break the code that assigns the base page template in my recent [NFC] code style compliance PR (*blush*). This PR fixes it.

Before
Throws "PHP Notice: Undefined variable: basepage_template"

After
No PHP Notice.

### DIFF
--- a/includes/civicrm.basepage.php
+++ b/includes/civicrm.basepage.php
@@ -622,13 +622,13 @@ class CiviCRM_For_WordPress_Basepage {
      *
      * @since 4.6
      *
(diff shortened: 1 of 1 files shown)


kcristiano:
@christianwach I missed this as well.   `r-run` works properly.

christianwach:
@kcristiano Thanks for merging. It happened because `civilint` was complaining about `apply_filters` inside an array so I moved it outside, then assigned the wrong variable as the param. Doh!

PR #231, opened 2020-12-18
Enhance CiviCRM's integration in WordPress

christianwach:
Overview
As originally proposed in the [Roadmap issue on Lab](https://lab.civicrm.org/dev/wordpress/-/issues/20). This PR is the result of reviews of the previous PR and responses to the development branch. Thanks to all those who took part.

Settings Page
Introduces a "Settings" page accessible via a sub-menu item for "CiviCRM" with quick access to settings that are of critical importance to those running CiviCRM under WordPress. Designed to be extensible via the [WordPress Meta Box framework](https://developer.wordpress.org/plugins/metadata/custom-meta-boxes/). Both the "Useful links" list and the "Maintenance tasks" list can be fully customised via filters.

Screen Shot 2020-12-18 at 14 05 55

This Settings Page is the one that I was referring to as the target for the migration of various settings from my CiviCRM Admin Utilities plugin. I guess it's obvious that without a page such as this, I can't migrate settings from CAU to here!

Integration Page
Introduces an "Integration" page accessible via a sub-menu item for "CiviCRM" with links to lots of resources that will be of interest to those running CiviCRM under WordPress. Also designed to be extensible via the WordPress Meta Box framework. Periodically brings in plugin data from the WordPress Plugin Directory API. It is also ready to bring in a curated list of other plugins from the CiviCRM "Community Messages" API when that's ready.

Screen Shot 2020-12-18 at 14 06 14

Now that there are no more hard-coded lists of resources, I think that this page is much more manageable.

"Quick Add Contact" Dashlet
Introduces a dashlet for the WordPress Dashboard that allows Contacts to be quickly added. Links to recently-added Contacts appear below the form (it only appears once they have been added) and persist for the duration of the session (to a maximum of 5 newly-added Contacts). The dashlet is positioned above the core "Quick Draft" dashlet for easy access.

Screen Shot 2020-12-22 at 20 58 05

Comments
Ready for review but not ready for merging yet because the route on `https://alert.civicrm.org/` does not yet exist.


kcristiano:
> Ready for review but not ready for merging yet because the route on https://alert.civicrm.org/ does not yet exist.

@christianwach any progress on integrating with alrt.civicrm.org ?  do we want to separate out the PR so we can merge the settings page?

kcristiano:
Specifically the Base page improvement is one I'd like to see in 5.34.

@totten any suggestions on how we can get theIntegration page mergable?    

These improvements are very much needed.  Specifically, the Base page improvement.  Since WP 5.5 the base page issues have grown exponentially.  In the past you could survive without a base page, now it's impossible.  The interface here is much better, understandable to users and in a Place WP user will find.   Users would never check CMS database integration, but they will look at settings.


I have done an `r-run` I have this running on my test sites and I'd merge now except for the WIP label due to integrating with alerts.civicrm.org

christianwach:
>> Ready for review but not ready for merging yet because the route on https://alert.civicrm.org/ does not yet exist.
>
> @christianwach any progress on integrating with alert.civicrm.org ? do we want to separate out the PR so we can merge the settings page?

@kcristiano The Settings page and Quick Add dashlet were meant to be enticements to get the route on https://alert.civicrm.org/ finalised. I guess it didn't work as expected 🤷‍♂️ 

FYI, MM thread here: https://chat.civicrm.org/civicrm/pl/83kw1gm9kjybifm1aefbubtmor

kcristiano:
Re-tested after `Disable "Integration Page" until Messages API becomes available` commit was added.

Did 3 `r-run`

- Buildkit of master
- Upgrade 5.33 to master with Patch applied
- Clean install of master with patch applied.

All works as expected.   New screen added, works as detailed above.

agileware-justin:
@christianwach @kcristiano love your work guys!

PR #232, opened 2020-12-30
Fix dev/core#2278 - pass contributionPageID as argument for PCP transact shortcodes

mattwire:
Overview
See https://lab.civicrm.org/dev/core/-/issues/2278

Before
Wrong contribution page ID passed.

After
Correct contribution page ID passed.

Technical Details
We do a simple lookup via the API for the contributionPageID linked to the personal campaign page.

Comments
@kcristiano @christianwach

This is a 5.33 backport and will need pushing to master too if approved.

### DIFF
--- a/includes/civicrm.shortcodes.php
+++ b/includes/civicrm.shortcodes.php
@@ -735,6 +735,10 @@ class CiviCRM_For_WordPress_Shortcodes {
           case 'transact':
             $args['q'] = 'civicrm/contribute/transact';
             $args['pcpId'] = $args['id'];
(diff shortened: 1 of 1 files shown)


kcristiano:
See https://lab.civicrm.org/dev/core/-/issues/2278#note_52280

I also did an `r-run` on opur test site and all 4 PCP pages work.  The test site now has 3 contrib pages and 4 PCP pages, so a better testing enviornment.

eileenmcnaughton:
Quick work @mattwire @kcristiano

PR #239, opened 2021-02-12
Better Shortcode handling

christianwach:
Overview
Discussion [on Lab](https://lab.civicrm.org/dev/wordpress/-/issues/90)

Before
Shortcodes that have had an "action" taken on them render as if they were on the Base Page.

After
Shortcodes that have had an "action" taken on them render in the context that they were embedded in.

Technical Details
If any action is taken via a Shortcode, a query string is appended to the URL and CiviCRM-WordPress goes into "Base Page Mode". [The relevant code that tries to sort all of this out can be found here](https://github.com/civicrm/civicrm-wordpress/blob/c40aa7106b21ebad25199e8c2b5d5b0e91d351b1/civicrm.php#L693-L732). This existing code tries to distinguish between "Base Page" and "Shortcode" contexts too early in the request lifecycle, so it never really does so reliably.

What this PR does is leave the `CiviCRM_For_WordPress_Basepage` and `CiviCRM_For_WordPress_Shortcodes` classes to sort out what's going on and act accordingly once the context has been discovered beyond doubt - i.e. once the query has been set up in the `wp` action.

Comments
I realise that the current behaviour may now be the "expected" behaviour so this may require plenty of eyes on it to tease out the consequences for designs in the wild.


includes/civicrm.shortcodes.php
```diff
@@ -295,20 +308,25 @@ public function render_single($atts) {
...
+      $args = $this->preprocess_atts($atts);
 
-    // invoke() requires environment variables to be set.
-    foreach ($args as $key => $value) {
-      if ($value !== NULL) {
-        set_query_var($key, $value);
-        $_REQUEST[$key] = $_GET[$key] = $value;
+      // Sanity check for improperly constructed shortcode.
+      if ($args === FALSE) {
+        return '<p>' . __('Do not know how to handle this shortcode.', 'civicrm') . '</p>';
```
MikeyMJCO, line 319: As the area's being touched this message, in the rare circumstances it gets hit isn't actually that useful.

Something like: "The shortcode could not be handled. It could be malformed or being used incorrectly." might be more useful :-)

MikeyMJCO:
This looks okay to me and testing on a BuildKit WPMaster build doesn't reveal any issues with vanilla Civi and out-of-the-box WP themes - which is a good start :-)

includes/civicrm.shortcodes.php
christianwach: @MikeyMJCO Done.

christianwach:
@MikeyMJCO Thanks for testing!

It's likely to be themes with page layout editors (e.g. Visual Composer) that run into the problem. The default themes rarely sub-divide the page content for layout purposes.

christianwach:
I think every built-in Shortcode is going to have to be tested since they have a variety of behaviours. Personal Campaign Pages, for example, are more complex than Profile forms.

agh1:
I just tried it out and this affects two situations:

1. You fail validation or otherwise do something that reloads the form.  This is the example in [your comment on the ticket](https://lab.civicrm.org/dev/wordpress/-/issues/90#note_54051).
2. You proceed through to the confirmation page and/or thank-you page.

I see absolutely no problem with the first situation.

For the second, however, I don't necessarily mind this in the abstract but it will almost certainly cause (minor, resolvable) problems for certain sites.  If a non-hijacked page has an event form on it and intro and/or footer text in the page, that text is probably only relevant for the first step in the registration process.  When that intro text appears at the top of the confirmation page and thank-you page, it will probably confuse the person filling out the form.

The same thing happens when you click the register now link from an event info page.

The workaround is generally simple: if you have information that's specific to the initial form, put it in the initial form rather than the page.  However, there certainly are plenty of sites that don't do that, and they would need to fix it when they upgrade.  Also, this would raise the question of what's the point of not hijacking the page if you need to clear everything out from the page.

christianwach:
@agh1 Thanks for reviewing, Andrew.

> If a non-hijacked page has an event form on it and intro and/or footer text in the page, that text is probably only relevant for the first step in the registration process. 

Yes, I agree, many sites could be built that way. But shouldn't that text more properly be defined in the "Online Registration" tab on the CiviEvent?

> The same thing happens when you click the register now link from an event info page.

The "text" I show in the screenshots on Lab is above and below the Shortcode in the WordPress Page _content_. That *could*, however, be anything - layout, e.g. other columns of content or similar; design e.g. graphics; a menu or whatever. Consider the following Page content - I've created the layout in the content the way that Visual Composer, Divi or Elementor would.

Here's the design of an info page:

Screenshot 2021-02-13 at 13 11 13

With this PR, the registration page retains the design:

Screenshot 2021-02-13 at 13 11 32

As does the Thank You page:

Screenshot 2021-02-13 at 13 44 28

Without this PR the layout changes to:

Screenshot 2021-02-13 at 13 13 45

> The workaround is generally simple: if you have information that's specific to the initial form, put it in the initial form rather than the page. However, there certainly are plenty of sites that don't do that, and they would need to fix it when they upgrade. 

Yes, I agree - I think this is more a question of documentation and managing expectations.

> Also, this would raise the question of what's the point of not hijacking the page if you need to clear everything out from the page.

Again, I agree. I think this actually makes the distinction between `hijack="0"` and `hijack="1"` clearer and more consistent. At present `hijack="0"` _actually means_ "do not hijack the page unless an action has been taken, in which case hijack the page".

agh1:
Yeah I guess it largely depends upon what we think of as content. Actual content is likely to be specific to the initial form only. However, for better or worse, lots of WP sites have page content that would be more like theming, strictly speaking, and should persist through the process.

This PR accommodates the latter at the expense of the former. It's probably the right call if we were starting from scratch, but we've got sites that are used to the former working and the latter failing.

Would it be possible to have a URL param that forces hijack? Even if nothing in this PR uses it, maybe an extension or form option could be set so that it adds the param as you proceed, making things work the old way if you want.

Organizations are rightfully sensitive about every change in donation forms, and as-is, this would force many to rework their site upon upgrade. Folks will miss the memo or not have the capacity to do it, and they'll be angry.

Feb 13, 2021 8:58:16 AM Christian Wach <notifications@github.com>:

christianwach:
> Organizations are rightfully sensitive about every change in donation forms, and as-is, this would force many to rework their site upon upgrade.

I guess there's another option - which would be to *add* another setting, e.g. `hijack="2"` perhaps labelled "True Hijack" or something. That would leave existing installs with their current behaviour whilst adding the functionality of this PR as an option for those who need it. Might be hard to explain, however.

christianwach:
> Actual content is likely to be specific to the initial form only.

I wonder for how many installs this is true? It would be interesting to find out, though I can imagine it being hard to do so.

My guess would be that the majority of Shortcodes are used with `hijack-"1"` where the main purpose of the Shortcode is to provide a more suitable URL for the CiviCRM content than the one provided by the Base Page. That's the only way I've ever used Shortcodes - but I'm aware that's anecdote not data.

kcristiano:
https://lab.civicrm.org/dev/wordpress/-/issues/90#note_54161

I've done and `r-run` and this is better merged than not.  As there is lively discussion I'll hold off merging for now.

agh1:
@christianwach to be clear I wasn't really suggesting an option to revert to precisely the status quo ("first page load").  I think few people like/expect that validation errors and coupon codes should cause hijack to commence.  Instead, it's really just the subsequent steps in the registration process, and CiviCRM uses URL params to identify which step you're on, so that seems like a good mechanism to support in `CiviCRM_For_WordPress_Basepage::register_hooks()`.  Basically, when you're deciding it's a basepage, immediately assume Yes if `civibasepage=1` or equivalent is in the request.  If this PR can support that, the rest could be handled by extension or independent core changes that would support a choice.

includes/civicrm.basepage.php
```diff
@@ -348,6 +332,17 @@ public function basepage_handler($wp) {
 
         global $post;
 
+        // Skip if this not the Base Page.
+        if ($basepage->ID !== $post->ID) {
```
agh1, line 336: This here seems to be the key change in behavior: instead of defaulting to basepage, this only goes to basepage mode if you're on the basepage as specified in the CiviCRM settings.  I imagine this is where to add something where the `if` is `FALSE` if there's a URL param specifying to use basepage mode.

Incidentally, I have no idea whether it's common to use another page as your basepage, but I wanted to flag the following change in behavior.  The examples below all assume that the default page `civicrm` exists and is the basepage set in CiviCRM.  Also, there is another page `other` that exists.

### Before (master)

| URL | Rendered |
| --- | -------- |
| http://example.org/civicrm/?civiwp=CiviCRM&q=civicrm/contribute/transact&reset=1&id=1 | CiviCRM Contribution page ID 1, template of `civicrm`, no page content |
| http://example.org/other/?civiwp=CiviCRM&q=civicrm/contribute/transact&reset=1&id=1 | CiviCRM Contribution page ID 1, template of `other`, no page content |

### After (with this PR)

| URL | Rendered |
| --- | -------- |
| http://example.org/civicrm/?civiwp=CiviCRM&q=civicrm/contribute/transact&reset=1&id=1 | CiviCRM Contribution page ID 1, template of `civicrm`, no page content |
| http://example.org/other/?civiwp=CiviCRM&q=civicrm/contribute/transact&reset=1&id=1 | The normal page content and template on `other`, with nothing from CiviCRM |

The only use case I've encountered of this is a site that had a special template for the conference section of their website and they wanted the registration form to have that look.  Of course, they could use a shortcode for this, instead.  Unlike the other concern about content on a shortcode page, I think this could be easily resolved with an upgrade message.

However, it might be good to run through some of the scenarios @christianwach and @kcristiano were thinking of in #17698 where the recorded basepage may not exist neatly in WP multisite.

christianwach: @agh1 Thanks for the analysis. 

* Yes, you're right; this PR does prevent rendering of CiviCRM content in the `/other` URL scenario (i.e. not on the Base Page) unless embedded via a Shortcode, of course. Personally, I think this kind of URL *should not* function, though I appreciate that people may be using it because it does.
* For the Base Page URL... I'd imagine that most Base Page requests are made using Clean URLs these days as it's been the default for quite some time. But as you point out, nothing changes for these Base Page URLs either way.
* I can't at the moment think of any impact when a Base Page doesn't exist - usually that means either (a) CiviCRM is incorrectly installed and it should have a Base Page or (b) it has no need for front-end content or (c) it's multisite and there's an appropriately configured install that doesn't need a Base Page. Happy to be proved wrong :-)

> `civibasepage=1`

What's your thinking here? How does this help?

christianwach:
> `civibasepage=1`

@agh1 What's your thinking here? Are you suggesting an additional param?

christianwach:
@agh1 Okay, so on further reflection, what I *think* you're saying is that you'd like certain scenarios (e.g. failed form validation) to remain in "Shortcode mode" whilst other scenarios should trigger "Base Page mode". Is that correct?

agh1:
> @agh1 What's your thinking here? Are you suggesting an additional param?

@christianwach yes, it would be an addtional param that, if truthy, would make it be in basepage mode.  Basically, instead of

```php
        // Skip if this not the Base Page.
        if ($basepage->ID !== $post->ID) {
```

do this

```php
        // Skip if this not the Base Page.
        if ($basepage->ID !== $post->ID && empty($_REQUEST['civibasepage'])) {
```

Basically I'm saying that the general behavior of this PR is great for most sites and it's not your responsibility to take care of all the folks with intro content above their shortcodes, but those people are out there and having something like this would make it possible to extend CiviCRM to simulate the status quo.

Nobody expects that failed validation should trigger base page mode, even though that happens to be the status quo.  A bunch of people expect that proceeding through registration/donation/etc. should trigger base page mode, since that's the status quo and semi-reasonable (even though what you're proposing is more reasonable).  I don't want to saddle you with dealing with those scenarios, but I don't want to leave folks high and dry.

A param is what seems like the most straightforward way to accomplish this, since the places we'd want to do it are proceeding through form steps that change params, but any other mechanism would also be fine.

agileware-justin:
> Nobody expects that failed validation should trigger base page mode, even though that happens to be the status quo.

Absolutely agree with this point.

> A bunch of people expect that proceeding through registration/donation/etc. should trigger base page mode, since that's the status quo and semi-reasonable (even though what you're proposing is more reasonable).

It would be a great improvement for CiviCRM shortcode to behave just like any other WordPress shortcode, just stay on the same page that it was embedded on. This enables you to have consistent Page Templates selected, rather than flipping between the CiviCRM base page and page where the shortcode was embedded. Keep the sidebars in place, menus etc. keeping the context of the page as is.

I am OK with changing the status quo here, because it's currently not great.

includes/civicrm.basepage.php
```diff
@@ -332,48 +332,58 @@ public function basepage_handler($wp) {
...
-         * the title using the 'wp_title' filter.
+        /**
+         * Skip if this not the Base Page, but allow "Base Page mode" to be forced.
+         *
+         * Return TRUE to force CiviCRM to render a Post of Page as if on the Base Page.
          *
-         * First, we try and set the title of the page object, which will work
-         * if the loop is not run subsequently and if there are no additional
-         * filters on the title.
+         * @since 5.35
```
kcristiano, line 340: as we are targeting master this should be 5.36

christianwach: Thanks for pointing this out @kcristiano I'll tweak when we've reached consensus 👍

christianwach:
> an additional param that, if truthy, would make it be in basepage mode

@agh1 Okay, so I've taken a more WordPress-y approach to this in my latest commit. I'm not keen on forcing the usage of a particular query var on developers - standard practice in these situations is to add a filter. You can now use any query var that you choose to implement in `$_REQUEST` in your callback to the `civicrm_force_basepage_mode` filter, e.g.

```php
add_filter( 'civicrm_force_basepage_mode', 'my_civicrm_force_basepage_mode', 10, 2 );
function my_civicrm_force_basepage_mode( $allow, $post ) {

	// Allow "Base Page mode" on a particular page or post.
	if ( $post->ID === $my_special_page->ID ) {
		return TRUE;
	}

	// Allow "Base Page mode" in the presence of a query var.
	if ( ! empty( $_REQUEST['civibasepage'] ) ) {
		return TRUE;
	}

	// Otherwise pass through.
	return $allow;

}
```

I think this provides greater flexibility.

includes/civicrm.basepage.php
christianwach: Fixed in 0b557e02a36a08a2b51d92e027874e196e8771c8

includes/civicrm.basepage.php
christianwach: @agh1 Does 659b8e432463945d68ba0205e694865c56553ea3 solve this for you?

agh1:
The latest change should work well.  I was hesitant at first because I was thinking it would be challenging to implement a per-event, per-contribution form, etc. override for this, but upon reflection it works better because the override could potentially be per-post, which is more specific to the issue.  (For example, maybe the donation form shortcode appears in two spots, and one should work like the new normal and retain page content while the other should work like the old normal and go into basepage mode as you proceed.)  And of course, a plugin could go check for the contribution page ID from the post and decide according to that.

At this point, this PR looks good and we just need a corresponding one for civicrm-core that adds an upgrade message explaining all of this.

includes/civicrm.basepage.php
agh1: @christianwach yep, see comment on the main thread!

kcristiano:
> At this point, this PR looks good and we just need a corresponding one for civicrm-core that adds an upgrade message explaining all of this.

I am not sure what needs to be detailed here.  This looks more like we are fixing unintended behavior.  I see upgrade notices as reporting changes that could break sites and that a user should notice and maybe delay the upgrade or make changes before upgrading.  I am not sure what action item we'd give the user.   

@agh1 Can you elaborate as to what I've missed?

BTW, I do think we need to update the docs for certain.  Just unsure of an upgrade notice.

agh1:
@kcristiano it's totally unintended behavior, but there are definitely people out there who have been putting their event/contribution intro content in the page.  I was thinking of something like this:

> The behavior of front-end CiviCRM forms inserted via shortcode in WordPress pages and posts has changed.  If the shortcode does not override page content, any page content surrounding the shortcode will now remain in place on every step of the form submission process--through to confirmation and thank you.  Content that is specific to the first step of the form should instead go in the introduction of the CiviCRM form.  Read more in [the documentation](https://docs.civicrm.org) for options if this is a problem.

The action item for most people is simply to move their intro text out of WP and into Civi.  The docs can explain that filter.

kcristiano:
OK.   I see your point.   No user will read the release notes, so we want a place to alert them on upgrade.

@christianwach what do you think?

agileware-justin:
3 thumbs up from me :+1:  :+1:  :+1:

christianwach:
I'm happy for there to be an upgrade notice.

PR #245, opened 2021-03-13
Fix access permissions for sub-pages and "Quick Add" dashlet

christianwach:
Overview
Fixes [this issue](https://lab.civicrm.org/dev/wordpress/-/issues/96) on Lab.

Before
Incorrect capabilities are assigned to Settings (and currently-disabled Integration) sub-pages and the "Quick Add" dashlet.

After
More appropriate capabilities are assigned to Settings (and currently-disabled Integration) sub-pages and the "Quick Add" dashlet.

Technical Details
As noted on the issue, `manage_options` seems a more sensible default capability than `administer_civicrm` for the sub-pages because `administer_civicrm` is not automatically granted to WordPress admins. I have added a filter for each page so that this can be overridden if required.

The "Quick Add" dashlet queries CiviCRM directly for the `add_contacts` permission because the permission may or may not have been granted to the WordPress User and therefore `current_user_can()` can't necessarily determine the User's status.

### DIFF
--- a/includes/admin-metaboxes/civicrm.metabox.contact.add.php
+++ b/includes/admin-metaboxes/civicrm.metabox.contact.add.php
@@ -69,6 +69,11 @@ class CiviCRM_For_WordPress_Admin_Metabox_Contact_Add {
    */
   public function register_hooks() {
 
+    // Bail if the current WordPress User cannot add Contacts.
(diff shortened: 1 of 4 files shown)


kcristiano:
I have done multiple `r-run` and this passes.

PR #247, opened 2021-03-26
Prevent pages and metabox from loading until CiviCRM is installed

christianwach:
Overview
The CiviCRM Settings and Integration pages and the Quick Add Metabox should not load unless CiviCRM has been installed. Follow up PR to #245

Before
Pages and Metabox load during installation.

After
Pages and Metabox do not load during installation.

Comments
The Quick Add Metabox was calling `CiviCRM_For_WordPress::initialize()` during the install process and causing an infinite redirect. This is fixed as a result of this PR.

Also applies the correct `@since` tag to the new methods.


kcristiano:
Thanks @christianwach This looks good.  passes `r-run`

PR #250, opened 2021-06-14
Added setup pcp page shortcode

kartik1000:
Overview
This PR has added a shortcode option to setup a PCP similar to other shortcodes present in PCP like info and transact. 

Before
_The current status. Please provide screenshots or gifs ([LICEcap](http://www.cockos.com/licecap/), [SilentCast](https://github.com/colinkeenan/silentcast)) where appropriate._

After
_What has been changed. Please provide screenshots or gifs ([LICEcap](http://www.cockos.com/licecap/), [SilentCast](https://github.com/colinkeenan/silentcast)) where appropriate._

Technical Details
see https://github.com/civicrm/civicrm-core/pull/20601

Comments
_Anything else you would like the reviewer to note_

### DIFF
--- a/includes/civicrm.shortcodes.php
+++ b/includes/civicrm.shortcodes.php
@@ -771,6 +771,15 @@ class CiviCRM_For_WordPress_Shortcodes {
             ]);
             break;
 
+          case 'setup':
+            $args['q'] = 'civicrm/contribute/campaign';
+            $args['action'] = 'add';
+            $args['component'] = 'contribute';
+            if ($args['id'] == '') {
+              $args['id'] = '1';
+            }
+            break;
+
           case 'info':
           default:
             $args['q'] = 'civicrm/pcp/info';
--- a/includes/civicrm.shortcodes.php
+++ b/includes/civicrm.shortcodes.php
@@ -753,7 +753,7 @@ class CiviCRM_For_WordPress_Shortcodes {
           $args['action'] = 'preview';
         }
 
-        switch($action) {
+        switch ($action) {
           case 'transact':
             $args['q'] = 'civicrm/contribute/transact';
             break;
--- a/includes/civicrm.shortcodes.php
+++ b/includes/civicrm.shortcodes.php
@@ -786,16 +786,6 @@ class CiviCRM_For_WordPress_Shortcodes {
             ]);
             break;
 
-          case 'setup':
-            $args['q'] = 'civicrm/contribute/campaign';
-            $args['action'] = 'add';
-            $args['component'] = 'contribute';
-            $args['id'] = civicrm_api3('Pcp', 'getvalue', [
-              'return' => 'page_id',
-              'id' => $args['id'],
-            ]);
-            break;
-
           case 'info':
           default:
             $args['q'] = 'civicrm/pcp/info';
--- a/includes/civicrm.shortcodes.php
+++ b/includes/civicrm.shortcodes.php
@@ -775,9 +775,10 @@ class CiviCRM_For_WordPress_Shortcodes {
             $args['q'] = 'civicrm/contribute/campaign';
             $args['action'] = 'add';
             $args['component'] = 'contribute';
-            if ($args['id'] == '') {
-              $args['id'] = '1';
-            }
+            $args['id'] = civicrm_api3('Pcp', 'getvalue', [
+              'return' => 'page_id',
+              'id' => $args['id'],
+            ]);
             break;
 
           case 'info':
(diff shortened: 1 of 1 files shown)


seamuslee001:
Jenkins add to whitelist

kcristiano:
This shortode looks for an existing PCP page and then returns the setup info.

The shortcode that is output is `[civicrm component="pcp" id="1" action="setup" mode="live" hijack="0"]`
The id here represents the PCP ID.

 `https://example.org/civicrm/contribute/campaign/?action=add&reset=1&pageId=5&component=contribute`  Is the CiviCRM page we'd want to see the shortcode for.   The Page ID is the Contribution Page, (in my example it'd be 5) by selecting the PCP page and inserting that ID we are trying to add a new PCP page to a PCP Page which fails.

I'd expect we need to choose the contribution page that has PCP enabled and then pass that Page id in.

kartik1000:
@kcristiano, I have replaced the value of id from PCP id with contribution page id, by fetching the same using `civicrm_api3`. Shall I add some error handling or alert message, in case the user does not select a PCP page and checks the option to setup a PCP? (We will not be able to fetch contribution page ID, without a valid PCP id). Thanks!

kcristiano:
If I create a new contribution Page with ID 5, with personal Campaign Pages enabled, we get a link that allows people to create their own pages on the PCP tab of the contribution page.  https://wpcvrc.tadpole.cc/civicrm/contribute/campaign/?action=add&reset=1&pageId=5&component=contribute  The Contribution  Page that this creates the PP on is https://wpcvrc.tadpole.cc/civicrm/contribute/transact/?reset=1&id=5

This form is based on the contribution page and creates a PCP campaign.   

image


The goal is to be able to add this form as a shortcode in WP.

If I have no existing PCP Pages this is what I get in the Modal:

image

If I have an existing PCP I can choose that, but it then will have the wrong params.

For setup, We would have to be able to select the base Contribution Page and then add that content via shortcode.

kartik1000:
@kcristiano Did you check after my commit? I think now we are fetching the contribution page id and then setting up a new pcp on top of that. I used the code from transact option to fetch the contribution page ID. The problem is if I don't select a PCP in first place, then their is no contribution id I can fetch,  but I guess then setting up a pcp from that shortcode does not make sense at all.

kcristiano:
@kartik1000 I did apply the latest commit when I tested and then the issue is when there is no PCP page yet there is nothing to select.

>  The problem is if I don't select a PCP in first place, then their is no contribution id I can fetch, but I guess then setting up a pcp from that shortcode does not make sense at all.

I understand and this makes sense.  The shortcode should be able to create the 'setup' of a PCP for any Contribution Page that has PCP enabled, regardless of whether  any PCP pages have been set up.

kartik1000:
@kcristiano got you, then will it make more sense to add this option under contribution pages category?

kcristiano:
@kartik1000 Perhaps this shortcode should be part of the Contribution Page shortcode:

image

We'd have to check and see if the PCP option was available per page.   But then we would get the proper page ID.

mattwire:
This sounds good. I don't know if the UI has the ability to filter by "contribution page has PCP enabled"? But maybe we could get around this by adding "(PCP enabled)" in the text for contribution pages that do and then checking that the "setup" page does not load if PCP is not enabled. It will rely a little bit on the user not making mistakes when setting up but I think this could be ok?

kcristiano:
That sounds like a good start.   I am also unsure if we can validate the PCP settings in the shortcode modal, so the user will have to take some responsibility.  We should do all we can to protect them, but in the end we do rely on them.

kartik1000:
@kcristiano, I have shifted setup PCP to contribution pages. Kindly have a look and let me know if it needs some changes.

kcristiano:
@kartik1000 I ahve the same issue.  If a contribution page allows PCP, but none are set up and I use `[civicrm component="contribution" id="1" action="setup" mode="live" hijack="0"]`  The page fails to load and bounces me to the home page.

If There is an existing PCP Page, the details of the ID load.  

I am testing on buildkit master.   To replicate, delete the PCP from `Help Support CiviCRM`

includes/civicrm.shortcodes.php
```diff
@@ -752,7 +752,22 @@ public function preprocess_atts($atts) {
...
           $args['action'] = 'preview';
         }
-        $args['q'] = 'civicrm/contribute/transact';
+
+        switch ($action) {
+          case 'transact':
+            $args['q'] = 'civicrm/contribute/transact';
+            break;
+
+          case 'setup':
```
kcristiano, line 761: The Direct URL is http://wpcvmaster.test/civicrm/contribute/campaign/?action=add&reset=1&pageId=1&component=contribute   The shortcode we get is `[civicrm component="contribution" id="1" action="setup" mode="live" hijack="0"]`  does the ID need to get mapped to `PageId`

mattwire: Fixed in #253

mattwire:
Replaced by #253

PR #253, opened 2021-07-17
Add setup pcp page shortcode

mattwire:
Overview
Replaces #250 

This PR has added a shortcode option to setup a PCP similar to other shortcodes present in PCP like info and transact.

Before
No shortcode

After
Shortcode for setup PCP

Technical Details

Comments

### DIFF
--- a/includes/civicrm.shortcodes.php
+++ b/includes/civicrm.shortcodes.php
@@ -321,7 +321,7 @@ class CiviCRM_For_WordPress_Shortcodes {
       // Check permission.
       $argdata = $this->civi->get_request_args();
       if (!$this->civi->users->check_permission($argdata['args'])) {
-        return $this->civi->users->get_permission_denied();;
+        return $this->civi->users->get_permission_denied();
       }
 
       // CMW: why do we need this? Nothing that follows uses it.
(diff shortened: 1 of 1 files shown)


kcristiano:
related https://github.com/civicrm/civicrm-core/pull/20601

kcristiano:
This looks good to set up the shortcode to enable the setup of PCP Pages on a WP Page.

Without the shortcode we need to use links that are in this format:   http://wpcvmaster.test/civicrm/contribute/campaign/?action=add&reset=1&pageId=1&component=contribute 

After the patch  we can add a shortcode and have a page such as : http://wpcvmaster.test/setup-pcp/

I've done a few `r-run` on Master and it works as expected.

There is one issue, but I believe it is not related to this code.

If you use the shortcode page the URL that the person creating the PCP page gets to distribute is http://wpcvmaster.test/setup-pcp/?civiwp=CiviCRM&q=civicrm/pcp/info/?reset=1&id=3  The 'clean form' is http://wpcvmaster.test/civicrm/pcp/info/?reset=1&id=3

Both links work.

Further when clicking the donate button on http://wpcvmaster.test/setup-pcp/?civiwp=CiviCRM&q=civicrm/pcp/info/?reset=1&id=3  we are taken to http://wpcvmaster.test/setup-pcp/?civiwp=CiviCRM&q=civicrm%2Fcontribute%2Ftransact&id=1&pcpId=3&reset=1  as opposed to http://wpcvmaster.test/civicrm/contribute/transact/?id=1&pcpId=3&reset=1

Again, both links work.   I expect we need to update the PCP template to allow for this use case and always return urls with the basepage.  

@christianwach any thoughts on the above.

I think this is OK to merge and we can address the URLs separately.

christianwach:
> always return urls with the basepage.

@kcristiano Not sure that's necessary - similarly to [the shortcode rendering issue](https://lab.civicrm.org/dev/wordpress/-/issues/90) I'd imagine that people would prefer links to point to the page where the shortcode is used - and thus render with the same design. But this is perhaps a tangential issue that needs to be addressed more generally.

I trust your `r-run` :)

kcristiano:
Thanks @christianwach I was wondering if I was overthinking.   

I agree this is good to merge, we'll need to get https://github.com/civicrm/civicrm-core/pull/20601  merged as well.  I'll make a not there.

PR #254, opened 2021-08-05
revert 240 Remove Base Page content filter once it has been use

kcristiano:
See https://lab.civicrm.org/dev/wordpress/-/issues/112 and https://lab.civicrm.org/dev/wordpress/-/issues/107

Getting redirect loop in some cases, and a confirmation page just showing the shortcoden in all others.

Suggest revert.

This should also be backported to ESR. 

ping @eileenmcnaughton

### DIFF
--- a/includes/civicrm.basepage.php
+++ b/includes/civicrm.basepage.php
@@ -522,9 +522,6 @@ class CiviCRM_For_WordPress_Basepage {
    */
   public function basepage_render() {
 
-    // We no longer need this filter, so remove it.
-    remove_filter('the_content', [$this, 'basepage_render']);
-
     // Hand back our base page markup.
     return $this->basepage_markup;
 


eileenmcnaughton:
@kcristiano - I take it you & @christianwach are in agreement on this (which is my criteria for merging wp prs)

eileenmcnaughton:
merged after confirming that @christianwach is away & can't weigh in in the timeframe we need but is generally aware of the issue & this makes sense as a quick response to a regression

agileware-justin:
See also https://lab.civicrm.org/dev/wordpress/-/issues/112#note_62843

rbaugh:
Tested on a couple of sites that we were having the issue on, both the Civi basepage and using a shortcode as well, and this change fixes the problems I was having in issue  [#112](https://lab.civicrm.org/dev/wordpress/-/issues/112).

PR #257, opened 2021-10-11
Fix unreplaced template vars in civicrm.settings.php when using wp-cli installer

demeritcowboy:
See also civicrm/civicrm-joomla#63

e.g. it leaves behind the literal string %%dbSSL%% in the DSN.


demeritcowboy:
Thanks!

PR #258, opened 2021-10-15
Introduce "Shortcode Display Mode" setting and functionality

christianwach:
Overview
Replaces #239. Discussion and screenshots [on Lab](https://lab.civicrm.org/dev/wordpress/-/issues/90).

Before
Shortcodes that had an "action" taken on them render as if they were on the Base Page.

After
Introduces a new metabox with a "Shortcode Display Mode" setting.

civicrm-shortcode-mode-metabox

Depending on the chosen setting:

* Shortcodes that had an "action" taken on them render as if they were on the Base Page.
* Shortcodes that had an "action" taken on them render in the context that they were embedded in.

Technical Details
Please refer to Lab and previous PR.

Comments
I think this PR avoids the need for documentation and for the upgrade notice in Core (which caused the previous PR to stall) because the default remains unchanged whilst "better Shortcode handling" can be implemented via the CiviCRM Settings screen for those who wish to use it.

### DIFF
--- a/assets/js/civicrm.options.js
+++ b/assets/js/civicrm.options.js
@@ -165,11 +165,14 @@
 
       // Assign properties.
       me.basepage_submit = $('#civicrm_basepage_submit');
+      me.shortcode_submit = $('#civicrm_shortcode_submit');
       me.email_submit = $('#civicrm_email_submit');
       me.cache_submit = $('#civicrm_cache_submit');
       me.basepage_select = $('#page_id');
+      me.shortcode_select = $('#shortcode_mode');
       me.email_select = $('#sync_email');
       me.basepage_selected = me.basepage_select.val();
+      me.shortcode_selected = me.shortcode_select.val();
       me.email_selected = me.email_select.val();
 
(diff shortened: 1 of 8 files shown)


christianwach:
> Instead, would it be possible to make `modern` the default and have an upgrade step that sets the site to `legacy`?

@agh1 I prefer the default being the status quo so that nobody's site breaks on upgrade. I think the default should be changed down the line when the setting is more widely known. No upgrade steps needed - I'm wary of upgrade steps because that's why the previous PR stalled.

christianwach:
Perhaps the availability of this new option can simply be flagged in the Release Notes for greater visibility at this stage?

I'm really keen for the underlying code to be merged for those that want/need it without causing issues for those that don't.

kcristiano:
Based on prior PR and comments here this looks good to merge.   I agree we should have it highlighted in the release notes.   We can also add a section to the docs https://docs.civicrm.org/sysadmin/en/latest/integration/wordpress/  unless you think this warrants a version specific notice in the upgrades section.

PR #259, opened 2021-10-16
Check recursively for CiviCRM Shortcodes in post content

christianwach:
Overview
Solves [this issue](https://lab.civicrm.org/dev/wordpress/-/issues/105) on Lab when Shortcodes are properly coded.

Before
CiviCRM Resources are not loaded for nested Shortcodes.

After
CiviCRM Resources are loaded for nested Shortcodes when those Shortcodes are properly coded. [This comment](https://lab.civicrm.org/dev/wordpress/-/issues/105#note_66401) explains the limitations.

### DIFF
--- a/includes/civicrm.shortcodes.php
+++ b/includes/civicrm.shortcodes.php
@@ -658,32 +658,45 @@ class CiviCRM_For_WordPress_Shortcodes {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


kcristiano:
I've done multiple `r-run` and this is an improvement and works as detailed.

PR #205, opened 2020-06-15
Create base page on every site on which CiviCRM is activated

christianwach:
Overview
----------------------------------------
Solves [this issue on Lab](https://lab.civicrm.org/dev/wordpress/-/issues/59).

Before
----------------------------------------
The base page is only created on main site in WordPress Multisite.

After
----------------------------------------
The base page created on every site on which CiviCRM is activated in WordPress Multisite.

Technical Details
----------------------------------------
This PR also introduces the `civicrm_basepage_main_site_only` filter which allow plugins to enable /restore the legacy "switch to the main site" behaviour.

PR #260, opened 2021-10-19
Streamline and enhance the plugin load procedure

christianwach:
Overview
This PR should solve a couple of issues on Lab:

* [404 Error when I use the Api v4...](https://lab.civicrm.org/dev/wordpress/-/issues/42)
* [Base page is only created on main site...](https://lab.civicrm.org/dev/wordpress/-/issues/59)

It replaces https://github.com/civicrm/civicrm-wordpress/pull/205 for the above issues.

It also streamlines the plugin load procedure(s):

* Adds the `anonymous_user` role and default capabilities on plugin activation.
* `wp civicrm install` now sets `wpLoadPhp` and creates the Base Page.

It should be noted that there is still work to be done when CiviCRM is configured in "Multi-Domain" mode. The procedures to set up new Domains are more complex than CiviCRM in "WordPress Multisite with multiple separate databases" mode. The Base Page, for example, cannot be identified until the new Domain has been properly set up - because the new CiviCRM Domain reads the "root" Domain's settings until that point.

Having said that, this PR improves the experience for most developers most of the time and should help future progress on streamlining "Multi-Domain" mode.


kcristiano:
This is the error I get 

```
PHP Fatal error:  Uncaught RuntimeException: Suite includes end-to-end tests (E2E\Api4\PermissionTest) which do not support CIVICRM_UF=UnitTests.
```

for:

`env CIVICRM_UF=UnitTests phpunit5 tests/phpunit/E2E/AllTests.php`

christianwach:
Hmm, hopefully 5d8bef7 fixes the failure.

kcristiano:
@totten Do you have any insights on hwy tests are failing on this PR?   I cannot seem to figure it out.

christianwach:
@kcristiano @totten I've made a couple of force-pushed changes to avoid what seem to be changes in the way that `civilint` now works - it was complaining about optional params before required params in method signatures, which it hasn't done before.

I'm still stumped why the tests are failing.

seamuslee001:
@christianwach that thing about the optional before required is actually not Civilint but php8, the test boxes now have php8 on as the PR testing version of php

christianwach:
Jenkins re test this please

christianwach:
Now that there's [a recent test success](https://github.com/civicrm/civicrm-wordpress/pull/264) to compare against, diffing the console output of this and #264 shows the failure being `4.6.36-setupsh.sql.bz2 ERROR`. 

@kcristiano @seamuslee001 @totten Anyone know what this means and how I can debug what's going on?

kcristiano:
That looks to like `wpLoadPhp` is failing to be set in the 4.6.36 upgrade test.  Based on when that setting was introduced, it should work and I am unsure why this fails:

```
[code] => -1
    [message] => DB Error: unknown error
    [mode] => 16
    [debug_info] => INSERT INTO civicrm_setting (`component_id`,`contact_id`,`created_date`,`created_id`,`domain_id`,`id`,`is_domain`,`name`,`value`) VALUES
(NULL,NULL,"20211221104414",NULL,"1",NULL,"1","wpLoadPhp","s:58:\"/home/jenkins/bknix-dfl/build/wp-260-48og8/web/wp-load.php\";")
 [nativecode=1364 ** Field 'group_name' doesn't have a default value]
    [type] => DB_Error
    [user_info] => INSERT INTO civicrm_setting (`component_id`,`contact_id`,`created_date`,`created_id`,`domain_id`,`id`,`is_domain`,`name`,`value`) VALUES
(NULL,NULL,"20211221104414",NULL,"1",NULL,"1","wpLoadPhp","s:58:\"/home/jenkins/bknix-dfl/build/wp-260-48og8/web/wp-load.php\";")
 [nativecode=1364 ** Field 'group_name' doesn't have a default value]
    [to_string] => [db_error: message="DB Error: unknown error" code=-1 mode=callback callback=CRM_Core_Error::exceptionHandler prefix="" info="INSERT INTO civicrm_setting (`component_id`,`contact_id`,`created_date`,`created_id`,`domain_id`,`id`,`is_domain`,`name`,`value`) VALUES
(NULL,NULL,"20211221104414",NULL,"1",NULL,"1","wpLoadPhp","s:58:\"/home/jenkins/bknix-dfl/build/wp-260-48og8/web/wp-load.php\";")
 [nativecode=1364 ** Field 'group_name' doesn't have a default value]"]
```

The setting is defined https://github.com/civicrm/civicrm-core/blame/0d0eca748e1bf13765cc6ade8cf7bd9ad5800557/settings/Core.setting.php#L905   and gropu is set to `core`

I am hoping others can get into more detail.

kcristiano:
related https://github.com/civicrm/civicrm-buildkit/pull/664

seamuslee001:
Jenkins re test this please

kcristiano:
Looks like the buildkit fix didn't help here - any way we can see why the 4.6.36 upgrade is failing?   I'd like to get past this as I had hoped we'd have this in 5.45 but we've missed that.

includes/civicrm.admin.php
```diff
@@ -774,24 +776,28 @@ public function admin_page_load() {
...
 
-    // Is it different to the one we've stored?
-    if ($setting !== $path) {
-      // Yes - set new path (this could be because we've changed server or location)
+    /*
+     * If we don't have a setting, create it. Also set it if it's different to
+     * what's stored. This could be because we've changed server or location.
+     */
+    if (empty($setting) || $setting !== $path) {
       CRM_Core_BAO_Setting::setItem($path, 'CiviCRM Preferences', 'wpLoadPhp');
```
seamuslee001, line 801: What happens if we change this to be 

```suggestion
      Civi::settings()->set('wpLoadPhp', $path);
```

totten:
> (@christianwach) Now that there's a recent test success to compare against, diffing the console output of this and #264 shows the failure being 4.6.36-setupsh.sql.bz2 ERROR.
>
> Anyone know what this means and how I can debug what's going on?

Not sure this was addressed directly; just to answer....

The test suite includes a series of database snapshots generated from historical releases (https://github.com/civicrm/civicrm-upgrade-test). IIRC, the PR selects [snapshots from 4.6.12 through current](https://github.com/civicrm/civicrm-buildkit/blob/master/bin/civi-test-run#L348); then, for each, it loads the DB and tries to run an upgrade. 

(*The kind of bug that it should call-out involves bootstrap-logic or upgrade-logic which relies on some new service or schema. For example, suppose v5.75.0 adds an entity `FooBar` and uses `FooBar` during bootstrap; it calls an API which calls a BAO which formulates a query `SELECT foo, bar FROM civicrm_foobar`. Now a user downloads 5.75.0 and tries to upgrade their DB from 4.6.12. Their 4.6 DB does not have `civicrm_foobar` - so bootstrap always fails - so the site becomes unusable, and it's not possible to run the upgrader.*)

To reproduce, I think you could manually load the snapshot DB and run `wp civicrm upgrade-db`. Similarly, `civicrm-upgrade-test.git` has a README/instructions for using the script. In `civibuild` environment, it's a bit more automated - any one of the following commands should reproduce (on a build named `wpmaster`):

```bash
## Run upgrade with a specific DB snapshot
civibuild upgrade-test wpmaster 4.6.36-setupsh.sql.bz2
civibuild upgrade-test wpmaster /path/to/buildkit/vendor/civicrm/upgrade-test/databases/4.6.36-setupsh.sql.bz2

## Run all DB snapshots between 4.6.12 and 4.7.0
civi-test-run -j /tmp/junit -b wpmaster upgrade@4.6.12..4.7.0
```

FWIW, I can confirm that I get a [similar failure](https://gist.github.com/totten/1cd88d644b60d8fd45105c84d3bbab3b) when running locally.

seamuslee001:
@totten my finding was that because we are now trying to set the setting before we start the upgrade that this if fails and as such https://github.com/civicrm/civicrm-core/blob/master/Civi/Core/SettingsBag.php#L419 group_name doesn't get applied to the insert SQL

kcristiano:
@totten Thanks for the anlaysis I'll try and look over the next few days.   

I've also determined that we had put a work around in buildkit https://github.com/civicrm/civicrm-buildkit/blame/master/app/config/wp-demo/install.sh#L60

However, this PR fixes this and that line causes a fatal error in te tests as far as I can see.   I was attempting to do a regex to add logic to only run if the version is less than or equal to 5.45, but so far my regex fails.   I can get the logic to not run that line if `$CIVI_VERSION -eq 'master'` to not run and that avoids the fatal error locally.  I'll try that with your above tests.

I'll also take a look at resolving the merge conflicts in wp-cli as I do want the clean up that was done in this PR.   

Any guidance on the regex to get a string version the is `le` a value would be appreciated.

totten:
For version comparisons in bash -- I've found it more effective to pass over to PHP's [`version_compare()`](https://php.net/version_compare) which handles multiple digits, alphas, betas, etc. There are a couple examples in `civibuild.lib.sh`, eg this creates a setting if version is <4.7.0:

```bash
  if cv ev 'exit(version_compare(CRM_Utils_System::version(), "4.7.0", "<") ?0:1);' ; then
    cv api setting.create versionCheck=0 debug=1
  fi
```

But that example uses `cv ev` and `CRM_Utils_System` - which may not work early-on. You can swap `cv ev` with `php -r`, but swapping `CRM_Utils_System::version()` is a little more tricky. I'd use the  `civicrm_get_ver` helper.

Perhaps something like this:

```bash
###############################################################################
## Compare the current Civi source-version to some reference-version.
##
## usage: civicrm_compare_ver 'OPERATOR' 'REFERENCE_VERSION'
## ex:    if civicrm_compare_ver '<' '5.20' ; then ... fi
## tip:   both inputs should be quoted, and neither input should have nested quotes
function civicrm_compare_ver() {
  local OP="$1"
  local REF="$2"
  REALVER=$(civicrm_get_ver "$CIVI_ROOT") \
    php -r 'exit(version_compare(getenv("REALVER"),"'$REF'","'$OP'")?0:1);'
  return $?
}

if civicrm_compare_ver '<' '5.20' ; then echo "Less than 5.20" ; else echo "More than 5.20" ; fi
if civicrm_compare_ver '>' '5.20' ; then echo "More than 5.20" ; else echo "Less than 5.20" ; fi
if civicrm_compare_ver '<' '5.50' ; then echo "Less than 5.50" ; else echo "More than 5.50" ; fi
if civicrm_compare_ver '>' '5.50' ; then echo "More than 5.50" ; else echo "Less than 5.50" ; fi
```

kcristiano:
@christianwach https://github.com/christianwach/civicrm-wordpress/pull/10  is my attempt to fix the merge conflicts

totten:
@christianwach @kcristiano I took a stab at rebasing in #271.

kcristiano:
@christianwach can we close this in favor of #271

kcristiano:
Merged https://github.com/civicrm/civicrm-wordpress/pull/271  which replaces this PR

PR #261, opened 2021-10-28
Prevent fatal error when the "the_title" filter has only one param

christianwach:
Overview
The "Give" plugin [calls "the_title" filter with only one param](https://github.com/impress-org/givewp/blob/775c659589fe106d06b20b9a5fa12a2b19ca9770/src/Views/IframeContentView.php#L102) and this produces a fatal error in `CiviCRM_For_WordPress_Shortcodes::get_title()`.

Before
Fatal error.

After
No fatal error.


seamuslee001:
@kcristiano any feelings on this?

kcristiano:
The code looks good, but I want to test on a clean install.   I expect I can merge shortly.

PR #263, opened 2021-12-02
Resolve conflict with AIOSEO stomping on CiviCRM Shortcode

christianwach:
Overview
Fixes [this issue on Stack Exchange](https://civicrm.stackexchange.com/questions/40765/wp-all-in-one-seo-plugin-conflict).

### DIFF
--- a/includes/civicrm.compat.php
+++ b/includes/civicrm.compat.php
@@ -71,6 +71,9 @@ class CiviCRM_For_WordPress_Compat {
     // Support Clean URLs when Polylang is active.
     add_action('civicrm_after_rewrite_rules', [$this, 'rewrite_rules_polylang'], 10, 2);
 
+    // Prevent AIOSEO from stomping on CiviCRM Shortcodes.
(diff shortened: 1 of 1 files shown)


kcristiano:
Unrelated test failure cc @seamuslee001 

Tested PR, Reproduced error without, apply PR Fixed.

Merging

arnaudbroes:
@christianwach did you ever manage to figure out why your shortcode broke when AIOSEO parsed it before you?

christianwach:
@arnaudbroes No sorry, I didn't have time to find the exact cause. My assumption is that it's because of a limitation of CiviCRM (which can only bootstrap once per HTTP request) meaning that it wouldn't render [the second time it was parsed](https://github.com/civicrm/civicrm-wordpress/blob/64b5f0752ec8be13afd65fefcda18a182aa86df9/includes/civicrm.shortcodes.php#L227). FWIW I don't think it's a problem with AIOSEO.

PR #265, opened 2021-12-21
(IGNORE) Dummy change

totten:
Dummy change to check current disposition of PR test runs.

PR #266, opened 2021-12-21
HookTest - Fix execution on PHP 8

totten:
@seamuslee001 This should address the failure demonstrated by #265.

Before
------

Test fails on PHP 8

```
CiviWP\HookTest::testFoo
Undefined array key "hook_was_called"

/home/jenkins/bknix-dfl/build/wp-265-47vsb/web/wp-content/plugins/civicrm/tests/phpunit/CiviWP/HookTest.php:22
```

After
-----

Test passes

Comments
--------

This is a superficial change - the assertion in question is supposed to validate the starting circumstance (asserting that the hook has not yet run). This is just a cleaner assertion that the hook has not yet run.


seamuslee001:
looks fine to me

PR #270, opened 2022-01-14
wp-cli/civicrm.php - Add 'civicrm pipe' subcommand

totten:
Add support for subcommand `wp civicrm pipe`

Complements https://github.com/civicrm/civicrm-core/pull/22262. Similar to https://github.com/civicrm/cv/pull/110.

Before
------------

Longer commands:

```bash
wp eval 'civicrm_initialize(); Civi::pipe();'
wp eval 'civicrm_initialize(); Civi::pipe("vlu");'
````

After
-------------

Shorter commands:

```bash
wp civicrm pipe
wp civicrm pipe vlu
```

Comment
--------------

If the command is successful, it will show a welcome/header line, e.g.

```javascript
{"Civi::pipe":{"v":"5.47.alpha1","l":["nologin"],"u":"untrusted"}}
```

You may then send JSON-RPC 2.0 requests, e.g.

```javascript
// Send request for `echo("hello world")`
{"jsonrpc":"2.0","method":"echo","params":["hello world"],"id":null}
// Receive response "hello world"
{"jsonrpc":"2.0","result":["hello world"],"id":null}
```

### DIFF
--- a/wp-cli/civicrm.php
+++ b/wp-cli/civicrm.php
@@ -47,6 +47,11 @@ if (!defined('CIVICRM_WPCLI_LOADED')) {
      * ===============
      * Run the CiviMember UpdateMembershipRecord cron (civicrm member-records).
      *
+     * wp civicrm pipe <connection-flags>
+     * ==================
+     * Start a Civi::pipe session (JSON-RPC 2.0)
+     * See https://docs.civicrm.org/dev/en/latest/framework/pipe#flags
+     *
      * wp civicrm process-mail-queue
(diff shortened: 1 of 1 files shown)


kcristiano:
@totten I think we should hold this until we rebase #260   That was only not merged due to test issues unrelated to the PR (as best I can tell) and with the recent installer wp-cli PR we have a large merge conflict to fix.

kcristiano:
Works as expected and documented.  `r-run` is successful.

PR #264, opened 2021-12-18
(dev/core#1615) wp-cli - Perform CLI installations using Civi\Setup

totten:
Overview
--------

Per https://lab.civicrm.org/dev/core/-/issues/1615, change how `wp civicrm install` performs installation.
Using the `Civi\Setup` means that:

* There's more shared installation code (eg WP GUI, Drupal GUI, cv CLI, and now WP CLI). Less duplicated installation code.
* Supports more install-time options (`$setup->getModel()`) to configure extensions, components, settings, etc
* Some options are impelmented more correctly (eg `$setup->getModel()->lang`).

CC @kcristiano @christianwach @colemanw 

Before
-------

* Loads the old installer, which (eg) does not have the ability to activate extensions or apply settings.

After
-----

* Loads `civicrm-setup` (aka `Civi\Setup`) which (a) has more functionality and (b) is used by the WP+Drupal GUIs and by `cv` CLI.

Technical details
-------------------

To test, here was my process:

* Setup an empty WordPress site and load the CiviCRM code (but don't run the installer yet).
    * Specifically, I used `wpmaster`. But I hacked `buildkit/app/wp-demo/install.sh` to skip Civi install. Then I ran `civibuild reinstall wpmaster`.
    * Alternatively, you might install a new WordPress site and download the tarball -- then apply the patches.
    * Alternatively, you might use `wpmaster` and uninstall Civi (deactivate plugin, delete `wp-content/uploads/civicrm`, drop SQL tables).
* Make a DB snapshot with the empty Civi+WP DBs (`civibuild snapshot wpmaster`)
* Run a series of commands like this:
   ```bash
    rm ~/bknix/build/wpmaster/web/wp-content/uploads/civicrm/ -rf
    civibuild restore wpmaster
    wp plugin activate civicrm
    wp civicrm install --dbhost=127.0.0.1:3307 --dbname=FIXME --dbuser=FIXME --dbpass=FIXME --lang=fr_FR
    ```

Comments
--------

This includes tangential changes in `--lang` handling:

* In `wp-cli/civicrm.php`, only require `--langtarfile` if the l10n data *is not available*. It's OK to use pre-existing l10n data.
* Use `Civi::setup()` for lang setup. This loads the localized data *and also* updates the initial `civicrm_setting`s to use the the language. (The prior routine neglected to set the active language.)

PR #271, opened 2022-01-14
Streamline and enhance the plugin load procedure (rebased)

totten:
(*This is a rebased version of @christianwach's #260. I've simply copied/pasted his original description. At the bottom, I'll add some comments about changes/conflict-resolution.*)

Overview
This PR should solve a couple of issues on Lab:

* [404 Error when I use the Api v4...](https://lab.civicrm.org/dev/wordpress/-/issues/42)
* [Base page is only created on main site...](https://lab.civicrm.org/dev/wordpress/-/issues/59)

It replaces https://github.com/civicrm/civicrm-wordpress/pull/205 for the above issues.

It also streamlines the plugin load procedure(s):

* Adds the `anonymous_user` role and default capabilities on plugin activation.
* `wp civicrm install` now sets `wpLoadPhp` and creates the Base Page.

It should be noted that there is still work to be done when CiviCRM is configured in "Multi-Domain" mode. The procedures to set up new Domains are more complex than CiviCRM in "WordPress Multisite with multiple separate databases" mode. The Base Page, for example, cannot be identified until the new Domain has been properly set up - because the new CiviCRM Domain reads the "root" Domain's settings until that point.

Having said that, this PR improves the experience for most developers most of the time and should help future progress on streamlining "Multi-Domain" mode.

Comments (Rebase/Conflict-Resolution)

There were two PRs going through contemporaneous review, #264 (which just changed the `wp civicrm install`) and #260 (which did code-cleanups and functional-updates across many files+functions... including `wp civicrm install`). Most files/functions/changes from #260 appear to merge cleanly with #264 - *except* for very harry conflicts around `wp civicrm install`.

So to make this PR, I did the following:

* Start from current `master` (which includes #264).
* Cherry-pick everything from #260 *except* for `wp civicrm install`.
* For `wp civicrm install`, read and manually reproduce the prose cleanups (*comments/messages; 3ad0227ba5ae3fad2cfecb1a4c69c8ac2a76df83*).

But there is one notable section that I did not reproduce:
 
Screen Shot 2022-01-14 at 1 45 32 PM

* This looks more functional. 
* I don't sufficiently understand the motivation/problem-scenario to reproduce/evaluate.
* For `r-run`ning `wp civicrm install`, I followed the same test procedure that I used in #264, and I don't see any obvious problems. The `wpLoadPhp` setting appears reasonable in my environment:
    ```
    $ cv ev 'return Civi::settings()->get("wpLoadPhp");'
    "/Users/totten/bknix/build/wpmaster/web/wp-load.php"
    ```
* It's possible that this last part is not needed any more -- ie it was drafted on top of the old installer; but the new installer is a separate implementation (and it's shared with the web-based install). eg Perhaps a prior fix aimed at web-based users is now sufficient for the CLI installer too.


totten:
I pushed up a patch which gets the 4.6.x upgrade-test to pass (1bdb8e9).

I spent some time running upgrades in different permutations (eg wrt to web-vs-cli as medium; wrt starting version; wrt ending version). With 1bdb8e9, it seems to be *as good as* the status quo -- ie

* __cli; 4.6.x => current master__: completes without error
* __cli; 4.6.x => current master plus this PR__: completes without error
* __web ui; 4.6.x => current master__: completes without error, but dashboard is weird (JS errors; doesn't show everything)
* __web ui; 4.6.x => current master plus this PR__: completes without error, but dashboard is weird (JS errors; doesn't show everything)

I haven't tracked where the issue with upgrading in the web ui originated -- but I can say these upgrades looked OK: 4.6=>5.13; 4.6=>5.27; 4.6=>5.39.

kcristiano:
@totten Thank you for reviewing this.   I'll do a new `r-run` and should be able to get it merged soon.

kcristiano:
I've done another review and `r-run` work as expected.

Merging.

totten:
Yay! Thanks @kcristiano

pboling:
For end users, what impact does this change have?  Does this mean we'll be able to upgrade the CiviCRM plugin via the normal plugin update feature of WordPress?

PR #272, opened 2022-02-07
Conditionally make CiviCRM's menu position an integer

christianwach:
Overview
Solves [this issue on Lab](https://lab.civicrm.org/dev/wordpress/-/issues/118) to prevent PHP notices in WordPress 6+.

### DIFF
--- a/includes/civicrm.admin.php
+++ b/includes/civicrm.admin.php
@@ -597,16 +597,27 @@ class CiviCRM_For_WordPress_Admin {
 
     $civilogo = file_get_contents(CIVICRM_PLUGIN_DIR . 'assets/images/civilogo.svg.b64');
 
+    global $wp_version;
+    if (version_compare($wp_version, '5.9.9999', '>')) {
+      $menu_position = 3;
+    }
+    else {
+      $menu_position = '3.904981';
+    }
+
     /**
      * Filter the position of the CiviCRM menu item.
      *
-     * Currently set to 3.9 + some random digits to reduce risk of conflict.
+     * As per the code above, the position was previously set to '3.904981' to
+     * reduce risk of conflicts. The position is now conditionally set depending
+     * on the version of WordPress.
      *
      * @since 4.4
+     * @since 5.47 Conditionally set because WordPress 6.0 enforces integers.
      *
-     * @param str The default menu position expressed as a float.
+     * @param str|int $menu_position The default menu position.
      */
-    $position = apply_filters('civicrm_menu_item_position', '3.904981');
+    $position = apply_filters('civicrm_menu_item_position', $menu_position);
 
     // Try and initialize CiviCRM.
     $success = $this->initialize();


kcristiano:
jenkins test this please

kcristiano:
@christianwach The code looks fine.  I wonder why the test fail but will test locally.

As 5.45 is ESR I do think this should be backported.  @totten what is the process for that?   

Once WP 6.0 ships all WP sites that don't have this will start throwing errors that the code is `doing_it_wrong`

christianwach:
> The code looks fine. I wonder why the test fail but will test locally.

@kcristiano Oh no, not this again. I've barely changed anything. Where's the :swear: emoji?

kcristiano:
* ([`-explain`](htps://doc.civicrm.org/dev/en/latest/standards/reviw/#r-explain)) PASS
* * ([`r-tes`](https://docs.civirm.org/dev/en/latest/standards/review/#r-test)) PASS
* * ([`r-code`](htps://docs.civicrm.org/dev/en/latest/standrd/review/#r-code)) PASS
* * ([`r-doc`](https://docs.civirm.org/dev/en/latest/stadards/review/#r-doc)) PASS
* * ([`r-maint`](https://docs.civicrm.org/dev/en/lateststandard/review/#r-maint)) PASS
* * ([`r-run`](htps://docs.civicrm.org/dev/enlatest/standards/review/#r-run)) PASS : No effect on WP versions 5.9.x, runs properly on 6.x
* * ([`r-user`](https://docs.civicrm.org/dev/n/latest/standars/review/#r-user)) PASS :  This should be backported as the WP version gets upgraded quickly and the CiviCRM version not so much.
* * ([`r-tech`](https://docs.civirm.org/dev/en/latest/standards/review/#r-tech)) PASS

We'll need to sort the tests and I think a backport to 5.45 ESR as well as 5.46 and RC is in order.

kcristiano:
@seamuslee001 It looks like the test is being flagged for CSFR 

`{"IP":"127.0.0.1","level":"security","referer":null,"reason":"CSRF suspected","is_error":1,"error_message":"SECURITY ALERT: Ajax requests can only be issued by javascript clients, eg. CRM.api3()."}`  

Was there a recent change on the E2E suite?

kcristiano:
@totten This does need to be in next weeks release and backported.   But as you commented on #273 these errors atrt with all WP PRs in 5.47 +

seamuslee001:
Jenkins re test this please

totten:
jenkins, test this please

kcristiano:
@totten I'd like to get this backported to 5.45 ESR.  The reasoning is that 5.45 will be around and supported when we need this patch.  Do I open up a PR here or do we have a different procedure for ESR?

totten:
@kcristiano Aaaah, I see! It took me a minute. (It's a little counter-intuitive for meb/c WP 6.0 hasn't even been released...) Let me verbalize just to make sure I follow correctly:

* Civi 5.45 ESR is supported through the end of July 2022. (https://lab.civicrm.org/esr/core/-/wikis/home)
* WP 6.0 is scheduled for April 2022. The current WP 5.9 will immediately go into a murky support status.
    > WordPress will be backported security updates when possible, but there are no guarantee and no timeframe for older releases. There are no fixed period of support nor Long Term Support (LTS) version such as Ubuntu's. None of these are safe to use, except the latest series, which is actively maintained. 
    > (https://codex.wordpress.org/Supported_Versions)
* The important timeframe is between April 2022 and July 2022.
* The patch needed to run Civi 5.45 with WP 6.0 is pretty conservative. There's no real downside.

So if I've understood correctly, then +1

> Do I open up a PR here or do we have a different procedure for ESR?

The procedure for ESR is designed to handle core vulnerabilities, so patches go through the private security repos, eg https://lab.civicrm.org/security/wordpress (in a branch `5.45-security`).

kcristiano:
Thanks @totten 

My experience is that WP will get updated quickly.  In fact WP now has automatic major upgrades !!  These are enabled by default and must be turned off!!

CiviCRM tends to be updated on a slower time table and it's done deliberatly and manually.

Therefore I expect ESR users to end up on WP 6.0 long before they go to ESR 5.51 (my guess on the next version for ESR).

I'll submit a backport PR for ESR.   As you say it's minor and not even needed until April.

PR #275, opened 2022-03-18
Fixes fatal error - function doesn't exist

stesi561:
As per https://developer.wordpress.org/reference/functions/is_favicon/ this function was only added in 5.4.0
meaning CiviCRM now breaks if installed on a site running Wordpress < 5.4.0

Overview
Checks function exists before calling.

Before
Fatal error

After
You should probably  upgrade your Wordpress site but if you are stranded on an old version due to plugin/theme incompatibilities at least you don't have a fatal error!

Technical Details

### DIFF
--- a/includes/civicrm.basepage.php
+++ b/includes/civicrm.basepage.php
@@ -322,7 +322,7 @@ class CiviCRM_For_WordPress_Basepage {
     }
 
     // Bail if this is a Favicon request.
-    if (is_favicon()) {
+    if (function_exists('is_favicon') && is_favicon()) {
       return;
     }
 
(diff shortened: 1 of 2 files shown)


kcristiano:
@christianwach any thoughts?  

Unless we increase the mon WP version this makes sense to me and looks good.

kcristiano:
`r-run`

Can reproduce on WP 5.3.   Fatal error on install.

Applying PR fixes the issue.

My want to switch branch and target RC.

stesi561:
Is that something that I need to do or is done by whomever is merging in?

christianwach:
Good catch @stesi561 

@kcristiano Yes this would be good to merge and backport if possible.

kcristiano:
targeting the RC branch as this is a regresion

PR #278, opened 2022-04-13
Always return boolean from initialize method

christianwach:
Overview
Makes sure `CiviCRM_For_WordPress_Admin::initialize()` always returns a boolean as declred in the docblock.

Before
`CiviCRM_For_WordPress_Admin::initialize()` currently returns a mix of `TRUE`, `FALSE` and `NULL`.

After
`CiviCRM_For_WordPress_Admin::initialize()` returns either `TRUE` or `FALSE`.


kcristiano:
Code looks good.  Better to be consistent.

`r-run` works as expected.

PR #279, opened 2022-07-04
Introduce "Permissions and Capabilities" metabox and functionality

christianwach:
Overview
Merges the CiviCRM-role-based functionality of the [CiviCRM Permissions Sync](https://develop.tadpole.cc/plugins/civicrm-permissions-sync) plugin into CiviCRM-WordPress.

Before
It was difficult to expose CiviCRM permissions as capabilities in WordPress without installing the [CiviCRM Permissions Sync](https://develop.tadpole.cc/plugins/civicrm-permissions-sync) plugin.

After
All CiviCRM permissions can be exposed as capabilities in WordPress via an optional custom role called "CiviCRM Admin".

In addition, the `civicrm_min_capabilities` filter runs when permissions are refreshed, which previously only ran when CiviCRM was activated, meaning any plugins which used that filter had to be installed and activated before CiviCRM itself.

civicrm-perms-metabox
The new metabox that allows the custom role to be enabled/disabled and capabilities/permissions to be refreshed.

### DIFF
--- a/assets/js/civicrm.options.js
+++ b/assets/js/civicrm.options.js
@@ -167,10 +167,12 @@
       me.basepage_submit = $('#civicrm_basepage_submit');
       me.shortcode_submit = $('#civicrm_shortcode_submit');
       me.email_submit = $('#civicrm_email_submit');
+      me.permissions_submit = $('#civicrm_permissions_submit');
       me.cache_submit = $('#civicrm_cache_submit');
       me.basepage_select = $('#page_id');
       me.shortcode_select = $('#shortcode_mode');
       me.email_select = $('#sync_email');
+      me.permissions_select = $('#permissions_role');
       me.basepage_selected = me.basepage_select.val();
       me.shortcode_selected = me.shortcode_select.val();
       me.email_selected = me.email_select.val();
@@ -359,6 +361,34 @@
       });
 
       /**
+       * Add a click event listener to the "Permissions and Capabilities" section submit button.
+       *
+       * @param {Object} event The event object.
+       */
+      me.permissions_submit.on('click', function(event) {
+
+        // Define vars.
+        var value = me.permissions_select.val(),
+            ajax_nonce = me.permissions_submit.data('security'),
+            refreshing = CiviCRM_Options_Settings.get_localisation('refreshing');
+
+        // Prevent form submission.
+        if (event.preventDefault) {
+          event.preventDefault();
+        }
+
+        // Modify button and select, then show spinner.
+        me.permissions_submit.val(refreshing);
+        me.permissions_submit.prop('disabled', true);
+        me.permissions_select.prop('disabled', true);
+        $(this).next('.spinner').css('visibility', 'visible');
+
+        // Submit request to server.
+        me.send('civicrm_refresh_permissions', value, ajax_nonce);
+
+      });
+
+      /**
        * Add a click event listener to the "Clear Caches" section submit button.
        *
        * @param {Object} event The event object.
@@ -453,6 +483,7 @@
       var saved = CiviCRM_Options_Settings.get_localisation('saved'),
           update = CiviCRM_Options_Settings.get_localisation('update'),
           clearing = CiviCRM_Options_Settings.get_localisation('clearing'),
+          refresh = CiviCRM_Options_Settings.get_localisation('refresh'),
           cache = CiviCRM_Options_Settings.get_localisation('cache');
 
       if (data.saved) {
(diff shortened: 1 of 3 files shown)


JoeMurray:
I think this is great functionality. Just wondering about policy of moving functionality out of core into extensions and whether it makes sense to make an exception here. There is a very useful functional improvement regarding refresh of perms. If this is not moved into core Wordpress Civi I imagine it would be more challenging to code. An alternative maybe good maybe bad would be to ship an improved extension with core WordPress CiviCRM. I'm more interested in there being explicit deliberation on approach than any particular outcome.

christianwach:
@JoeMurray I'm not entirely clear what you mean by "to ship an improved extension with core WordPress CiviCRM". This plugin (assuming that's what you're referring to by "core WordPress CiviCRM") doesn't have "extensions" as such.

This functionality in this PR used to be in [a plugin](https://develop.tadpole.cc/plugins/civicrm-permissions-sync) but that had its limitations because of the way the `civicrm_min_capabilities` filter currently functions. Moreover it needed to be installed and active before CiviCRM itself. And the custom role is entirely optional.

Could you expand on what an alternative approach might look like in terms of its architecture?

mattwire:
@christianwach I'm pleased to see this PR. In my opinion it's pretty essential functionality for a successful, permissioned, WordPress and CiviCRM integration and belongs in this civicrm-wordpress module

christianwach:
@mattwire Thanks Matt, I agree :)

BTW, if you use the [Groups plugin](https://wordpress.org/plugins/groups/) for managing WordPress capabilities (I made [CiviCRM Member Sync](https://wordpress.org/plugins/civicrm-wp-member-sync/) compatible with Groups a long time ago) then there's a rewrite of [CiviCRM Groups Sync](https://develop.tadpole.cc/plugins/civicrm-groups-sync) on the WPCV GitHub account called [Integrate CiviCRM with Groups](https://github.com/WPCV/wpcv-civicrm-groups-integration) that works with this PR to refresh CiviCRM capabilities to the Groups plugin's list of capabilities. I ported the "other half" the of CiviCRM Permissions Sync functionality there.

kcristiano:
@JoeMurray I don't see this as a new feature or enhancement that LEXIM would cover.      I see this as a gap for WP users best fulfilled as a WP only option.  

The code looks good to me and I'd rather see this in early for more extensive testing.   @mattwire do you agree?

JoeMurray:
It's rare to change an extension into core code but I think this is a reasonable move in the circumstances. @eileenmcnaughton do you agree it makes sense here?

eileenmcnaughton:
@JoeMurray I don't thnk there is a hard and fast rule about not bringing extension functionality into core -- at some we probably will bring in the `lineitemeditor` .... it's probably more about code maintainability & appropriateness

christianwach:
Hi all... I don't know about others, but the way I name things is:

* CiviCRM-Core = The code in [the core repo](https://github.com/civicrm/civicrm-core).
* CiviCRM-WordPress = The code in this repo.
* Extension = A civix-based CiviCRM Extension.
* Module = A Drupal add-on.
* Plugin = A WordPress add-on.

I realise it's a little fluid because CiviCRM-Core and CiviCRM-WordPress are merged to become a plugin, but I find these definitions help me.

@JoeMurray Per the above, [CiviCRM Permissions Sync](https://develop.tadpole.cc/plugins/civicrm-permissions-sync) is a plugin some of whose functionality is being brought into CiviCRM-WordPress. (The rest went into [Integrate CiviCRM with Groups](https://github.com/WPCV/wpcv-civicrm-groups-integration).)

There is precedent for this:

* CiviCRM-WordPress has benefitted from lots of code that I have ported from [CiviCRM Admin Utilities](https://github.com/christianwach/civicrm-admin-utilities). 
* Andrei's [CiviCRM WP REST API Wrapper](https://github.com/mecachisenros/civicrm-wp-rest) plugin was merged into CiviCRM-WordPress. (FWIW, I will be upgrading this to include API4 support.)

The criteria (for me at least) are:

* Discoverability: I doubt many people were aware of the existence of CiviCRM Permissions Sync.
* Essential functionality: Without CiviCRM Permissions Sync, CiviCRM does not interface with WordPress in an expected manner.

I expect there are other reasons but that's my tuppence worth :)

JoeMurray:
I am satisfied that there has been a good discussion and decision in favour of moving extension functionality into core in this case. I leave other points of review to others. Thanks for your work here Christian.

eileenmcnaughton:
I terms of review - I'm kinda assuming you are going to cover that @kcristiano ....

kcristiano:
@christianwach 

In further testing --

I have a site with the `civicrm_admin` role.

If I do the following:

image

The role is removed and users in the role now have no role.  

Is this intended?   I can see why we may want to allow removing the role, but if we do shouldn't we detail that in the meta box?   let me know.

We can add something similar to:

image

PR #283, opened 2022-10-27
Add WPML integration for proper CiviCRM language generation #133

shaneonabike:
Overview
Fixed [dev/wordpress#133](https://lab.civicrm.org/dev/wordpress/-/issues/133)

Before

After
Translated pages are being shown. :fireworks:


kcristiano:
test this please

mlutfy:
jenkins, test this please

shaneonabike:
> Reading through the code, it looks like you've inserted this inside of [this method](https://github.com/civicrm/civicrm-wordpress/blob/master/includes/civicrm.compat.php#L87) and removed support for polylang.

If you take a look at the diff yuou'll see that I injected two new WPML functions (well three but one is a helper).

+ ```rewrite_rules_wpml```
+ ```rewrite_civicrm_urls_wpml```

The only change that I made for polylang is that I moved the code that registers the appropriate collected_rewrites into an appropriate rule and then flushes the rules if necessary. I figured there was no reason to duplicate this code since I imagine we want to reduce code glut. You can see that moved code in ```wp_add_rewrite_rules```

> 
> If we want to add another supported WP multiple language plugin, we should be it's own method so we don't conflict with what we already working.

For sure! You will see that above or in the diff.

> 
> Can you also detail how to test this and include the settings needed for WPML?

I outlined some test scenarios in [dev/core/133](https://lab.civicrm.org/dev/wordpress/-/issues/133), but did you need more or you need the tests to be outlined here. If you could explain more that would be helpful.

shaneonabike:
jenkins, test this please

seamuslee001:
Jenkins add to whitelist

kcristiano:
@shaneonabike Removing this section is the issue:

https://github.com/civicrm/civicrm-wordpress/pull/283/files#diff-6b01105839eb0dbea401150ea50124f23d512ba16f41671c5337134ea1127499L119

I would like a bit more details on the issue as well as away to test.  WPML is not open source and I have no access to the code.

Do you also have any comments on how ongoing support will be handled?  Once we merge this in we have an obligation to maintain, and that is tough on a plugin that has a paywall.

shaneonabike:
@kcristiano That code got refactored into ```wp_add_rewrite_rules```, which is then called by the WPML function and the Polylang function. I figured that if I had duplicated the code people would have asked me to create a helper function. 

I didn't remove any code just refactored it into a function is all. Does that make sense? If not I'm happy to jump on Mattermost to explain more...

## Testing

+ Install WPML plugiin
Selection_019

1. Translate the CiviCRM Page
Selection_020
6. Translate said page above
7. View live version of page

> Why WPML

> Long-term Support

shaneonabike:
PS - I'll fix the compatibility stuff next week since I have a few client priorities first

christianwach:
> I didn't remove any code just refactored it into a function is all.

I agree with @kcristiano - the Polylang "Base Page ID per language" is no longer retrieved via `pll_get_post()`.

It would make it easier to review this if there was:

* One commit with the working WPML code added.
* A second (if necessary) which extracts common code into a helper method.

christianwach:
@shaneonabike I thought I'd add a note on the general approach you're taking in supporting WPML. WPML recommends [using `apply_filters()`](https://wpml.org/documentation/support/wpml-coding-api/) to interact with its API. So, for example:

* Instead of [reading the WPML settings directly](https://github.com/civicrm/civicrm-wordpress/blob/9bf13b1d1fa3743aaad906fe4b6ebf7c8010b5e5/includes/civicrm.compat.php#L154), one should use the [`wpml_setting` filter](https://wpml.org/wpml-hook/wpml_setting/).
* For [reading the active languages](https://github.com/civicrm/civicrm-wordpress/blob/9bf13b1d1fa3743aaad906fe4b6ebf7c8010b5e5/includes/civicrm.compat.php#L155), one should use the [`wpml_active_languages` filter](https://wpml.org/wpml-hook/wpml_active_languages/). 

Using the WPML API in such a manner will respect any other plugins that hook into those filters but also avoids needing to touch the WPML internals. From a compatibility standpoint, it would be preferable if there was much greater use of the recommended WPML API approach than the current PR contains.

shaneonabike:
Hi @christianwach yes you're right I should use them 100%. I had used them in some cases, but it's true that this is the format of WPML. I'll change that around.

@kcristiano oh man I think when I merged my changes into my gitlab version it removed some of the polylang code. I really don't understand how that happened but now I see what you are talking about. So sorry and thanks for being clearer. In my local version I copied the polylang code completely. If we decide to use the helper function after (which I'm using for WPML) then we can since it is duplicating code.

I still have to rework all the Jenkin issues before submitting again. I will also test everything to make sure the suggestions of @christianwach are working too.

Thanks so much for your thorough feedback. :ghost:

shaneonabike:
After some serious thought, and discussion with @mlutfy we were thinking that perhaps I should build this into an extension that ties into CiviCRM via a Wordpress extension. We could make this available in all the available channels and via the blog.

I also think that it might be useful to think about doing this with polylang. Presently, the filter calls are being made regardless if Polylang is installed, and then certain checks are made. Whereas, if we migrated this to an plugin it could be enabled if the person is using Polylang and not otherwise.

It would also mean no means of automated testing and if it breaks we are going to know through our clients and repair it.

What do you think @christianwach @kcristiano ?

PS - I'm going to take all the suggestions that were made about using the ```apply_filters``` and modify my code to use it appropriately so that there is no issue there, but I think I'll migrate this to an plugin.

shaneonabike:
jenkins, test this please

shaneonabike:
@christianwach I removed the helper function for now, and applied the ```apply_filters```. This was a good suggestion as that is the standard method in WP.

@kcristiano you can now test this out if you had already applied for a developer account for WPML

Let me know if there are any other changes. I have tested this out in the three scenarios that WPML provides so far and it's working great.

shaneonabike:
jenkins, test please

shaneonabike:
jenkins, test please

kcristiano:
@shaneonabike I never heard back from WPML on a developer license :(   Can you also look and see about the test failure?

kcristiano:
This test failure looks valid - https://test.civicrm.org/job/CiviCRM-WordPress-PR/280/testReport/junit/(root)/E2E_Core_PathUrlTest/testGetUrl_WpAdmin/   The other is a curl timeout that I am less concerned about

shaneonabike:
@kcristiano I contacted WPML (since I have an account already) and they wrote me back. Apparently @christianwach received a developer license. Perhaps they can send that to you as well to share?

> In fact, a fellow by the name of Christian Wach reached me at the end of October last year to explain the same. Is he still working on the project?
> I provided him with a WPML subscription for access to our latest files as well as some relevant documentation that may be helpful for creating the integration. 
> I have not received any further updates from him since then. 

I'll take a look at the issues it's raising and see if I can't figure out what is happening

christianwach:
@shaneonabike Yes that's right, they revived an ancient WPML account that I had - but in the absence of any resources to look into this further, I let the matter drop.

shaneonabike:
I'm going to reforce a test because the new code I introduced won't get activated unless WPML is installed so technically it doesn't mess with any page settings. There was a curl timeout so let's check that. I'll also test on our own suite just to be sure.

@christianwach fair enough. I wasn't attempting to blame, but wondering if you would be willing to share with @kcristiano since I believe they are motivated in testing this further. I don't think WPML will provide an additional free license from my conversation with them.

jenkins, test please

shaneonabike:
jenkins, test please

christianwach:
@shaneonabike Thanks for making your code additions much more visible. I'll speak to @kcristiano about how to move forward with this. In the meantime, if you could rebase/squash this so that a single commit remains, that would be helpful.

mlutfy:
jenkins, test this please

christianwach:
@shaneonabike Sure, no problem.

PR #285, opened 2022-11-25
CiviMail - BAO files were renamed in core

colemanw:
Overview
See https://github.com/civicrm/civicrm-core/pull/24982


kcristiano:
Merging as core PR is merged and better to test early.

PR #289, opened 2023-02-13
Add WPML integration for proper CiviCRM language generation #133

shaneonabike:
Overview
Fixed [dev/wordpress#133](https://lab.civicrm.org/dev/wordpress/-/issues/133)

Before

After
Translated pages are being shown. :fireworks:

### DIFF
--- a/includes/civicrm.compat.php
+++ b/includes/civicrm.compat.php
@@ -71,6 +71,10 @@ class CiviCRM_For_WordPress_Compat {
     // Support Clean URLs when Polylang is active.
     add_action('civicrm_after_rewrite_rules', [$this, 'rewrite_rules_polylang'], 10, 2);
 
+    // Support Clean URLs when WPML is active.
+    add_action('civicrm_after_rewrite_rules', [$this, 'rewrite_rules_wpml'], 10, 2);
+    add_filter('icl_ls_languages', [$this, 'rewrite_civicrm_urls_wpml']);
+
     // Prevent AIOSEO from stomping on CiviCRM Shortcodes.
     add_filter('aioseo_conflicting_shortcodes', [$this, 'aioseo_resolve_conflict']);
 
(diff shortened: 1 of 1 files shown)


shaneonabike:
## Testing

+ Install WPML plugiin
Selection_019

1. Translate the CiviCRM Page
Selection_020
6. Translate said page above
7. View live version of page

> Why WPML

> Long-term Support

shaneonabike:
jenkins, test this please

shaneonabike:
Hi @christianwach 

I decided to try and debug the issue that is failing here, but when running the Jenkins tests on our own servers it's not failing with 5.58.0. I get ```ok 1409 - E2E\Core\PathUrlTest::testGetUrl_WpAdmin```. I feel that it's a red herring that this is failing.

I specifically added a test to see if WPML is installed on line [166](https://github.com/civicrm/civicrm-wordpress/pull/289/files#diff-6b01105839eb0dbea401150ea50124f23d512ba16f41671c5337134ea1127499R167) to eliminate it from changing anything unless the proper conditions are met.

Then line line [77](https://github.com/civicrm/civicrm-wordpress/pull/289/files#diff-6b01105839eb0dbea401150ea50124f23d512ba16f41671c5337134ea1127499R76) shouldn't get fired because the filter wouldn't normally exist unless WPML is installed also.

Just to be 100% sure the site that I am presently using this patch on I ran ```cv url civicrm/contribute?reset=1``` and it generates ```https://mysite.org/wp-admin/admin.php?page=CiviCRM&q=civicrm%2Fcontribute&reset=1```, which is correct and accurate like our test.

Is there some reason Jenkins might be failing here?

shaneonabike:
Hi @christianwach @christianwach 

Please let me know if there is anything else that I can do to move this along. I'm willing to help in anyway and would love to get this put into the next major release if we can. With that being said I have no clue what your workloads are.

christianwach:
@shaneonabike I followed the WPML setup instructions and seem to see valid translations without this PR. Here's German, for example:

Screenshot 2023-03-01 at 09 56 50

This is content rendered at `https://example.org/de/civicrm/event/info/?reset=1&id=2`. Same successful result for French.

So does this PR solve just the problem of the Language Switcher menu not containing the correct alternative URLs?

christianwach:
FYI, with you PR, my logs are filling up with:

```
[01-Mar-2023 10:10:14 Europe/London] PHP Warning:  Undefined variable $wpml_lang_conf in /path/to/httpdocs/wp-content/plugins/civicrm/includes/civicrm.compat.php on line 271
[01-Mar-2023 10:10:14 UTC] PHP Warning:  Undefined variable $wpml_lang_conf in /path/to/httpdocs/wp-content/plugins/civicrm/includes/civicrm.compat.php on line 271
```

I also note that your PR doesn't include a solution for different domains per language. Is this because you think it's an edge case?

christianwach:
When I set WPML to "Language name added as a parameter", my URLs in CiviCRM become of the form:

```
https://example.org/civicrm/?lang=fr/event/info/?reset=1&id=3
```

These obviously fail to render properly. It looks like you'll need a PR in CiviCRM Core to go alongside this one.

christianwach:
When I set WPML to "Different languages in directories" and "Use directory for default language", then on the "Manage Events" screen (regardless of the language I choose for WordPress) all my URLs in the "Event Links" dropdown become of the form:

```
https://example.org/fr/civicrm/event/info/?reset=1&id=3
```

They're always in French. They seem to show properly on individual Event "Info & Settings" screens, however.

christianwach:
Okay, that's enough for me for today...

@shaneonabike It appears that this PR has a lot of issues that need addressing. I'll check back when you think you've got this PR to a stage where you think WPML integration is more stable.

shaneonabike:
> So does this PR solve just the problem of the Language Switcher menu not containing the correct alternative URLs?

That's correct. The languages work but the language switcher does not. So what this patch does is register the URLs properly with WPML, and then provide the alternative languages via the language switcher.

> FYI, with you PR, my logs are filling up with

When I refactored I had missed one section for retrieving the actual ```language_setting``` so this is why it was complaining and actually failing.

> When I set WPML to "Different languages in directories" and "Use directory for default language", then on the "Manage Events" screen (regardless of the language I choose for WordPress) all my URLs in the "Event Links" dropdown become of the form

At first I wasn't able to reproduce this, but now I see what is happening. Essentially, in the backend it is producing links of the opposite language. Oddly enough, when I try this with my settings (having English set in my WP profile) the links are in English.

I'll take a look at this one and try and resolve it although it seems odd to me that it would produce the wrong links.

> When I set WPML to "Language name added as a parameter"

Can you try this again with the latest. The issue you were having should be resolved by the code that was failing, which has been fixed [here](https://github.com/civicrm/civicrm-wordpress/compare/master...shaneonabike:civicrm-wordpress:wpml-integration-2#diff-6b01105839eb0dbea401150ea50124f23d512ba16f41671c5337134ea1127499R271-R281). I'm sorry for missing that one.

I will also try on my end, because when I had tested this it seemed ok

I rebased everything as well so there is only one patch to apply.

shaneonabike:
> When I set WPML to "Different languages in directories" and "Use directory for default language", then on the "Manage Events" screen (regardless of the language I choose for WordPress) all my URLs in the "Event Links" dropdown become of the form

I was really stumped by this issue, because I couldn't figure out how it related to my code. Finally, I realized that there is another place that handles language negotiation for CiviCRM. [This](https://github.com/civicrm/civicrm-core/blob/12c48f7ee72450703cb8b64b13d6573b0808a127/CRM/Utils/System/WordPress.php#L640) chunk of code determines the locale for the backend. But in my tests the solution for WPML is not actually the correct one. Active languages only provides languages activated in WPML not actual "active" or user "selected" language.

According to WPML [docs](https://wpml.org/documentation/getting-started-guide/language-setup/setting-the-wordpress-administration-language/) the "Admin language is independent of the Site's Front-end Language". It's selected directly from the user's profile.

So the million dollar question is whether you want me to roll this fix into patch or a new one. The issue you pointed out here is an actual bug present in CiviCRM. I can post another one because I already have the solution i just need to understand how you think I should proceed.

All the other issues you pointed out should be working now, and I think this is pretty ready for prime time unless you find any other little fires or fixes I need to put out.

cc @christianwach @kcristiano

christianwach:
@shaneonabike I would fix CiviCRM Core before returning to this PR.

shaneonabike:
Hi @christianwach @kcristiano 

Upon further investigation, I realized that the issue in core only pertains to whether the proper language locale details are loaded in the front-end properly.

Second: My patch here was just to make the front-end integrate better to WPML and therefore switch the language properly (ie. viewing an Event in one language and then using the language switcher to change languages).

Lastly: I have to come across an instance of polylang to verify, but going through the code it doesn't seem that ```Civi/Core/Paths.php``` actually uses the ```$locale``` of the user at all in generating the links. The link in question on backend Event [links](https://github.com/civicrm/civicrm-wordpress/pull/289#issuecomment-1449920205) is generated by the SMART crmURL, which in turns goes back to getURL in the Wordpress.php. But in backtracing I don't see how this ever really uses the ```$locale``` of a user.

Maybe I'm missing something here, because it does go all over the place. But I just wanted to be sure we weren't holding this patch up due to an issue that also might be present for Polylang.

shaneonabike:
I guess what I am asking is whether anyone can test this on Polylang to see if the same thing happens? This would really help me determine if there is something missing in the patch or this is a generalized problem.

I noticed in the Drupal they implement the function ```public function languageNegotiationURL``` which provides the proper URL for each language, but WP implementation does not. I don't know if this is a lead, but maybe.

Do either of you @christianwach @kcristiano have time to double check a polylang install to see if you can reproduce the issue of events link for me?

shaneonabike:
Hi @christianwach @kcristiano 

I was wondering if you had a chance to check out Polylang to see if it's an issue there? I would <3 to commit this patch as I think it resolves a base of issues for people using WPML. At the same time, I recognize we should do this right. So I would love to determine if the language issue I found above relates to both extensions or just WPML.

Thanks for your time

christianwach:
> Thanks for your time

@shaneonabike This is, unfortunately, the problem... because I don't have any clients who require this, I am not able to devote the significant investment of time needed to resolve this.

kcristiano:
test this please

christianwach:
@kcristiano This PR will need rebuilding in the new compatibility architecture. @mlutfy [says that this has been done](https://github.com/civicrm/civicrm-core/pull/27128#issuecomment-1710642027) so I assume there will be an update or replacement PR in due course.

mlutfy:
I opened #303 to forward-merge changes on the 5.66 branch (polylang), then we can update this PR.

shaneonabike:
@christianwach @kcristiano As @mlutfy mentioned my plan was to integrate the changes I made here directly into the new framework. Then we should be able to move forward with testing, and then eventually :crossed_fingers: integration.

shaneonabike:
I'm going to close this issue for now and create a new one. After the work of @christianwach and the refactoring that was done :fireworks: I see that a bunch of changes that I made were integrated. There is a few things missing to make it fully completed so I'll create a patch just to handle that.

@christianwach @kcristiano @mlutfy

christianwach:
@shaneonabike Good call. Hopefully the new code simplifies your task.

PR #292, opened 2023-02-28
Use custom actions for rendering admin pages

christianwach:
Overview
Fixes [this issue on Lab](https://lab.civicrm.org/dev/wordpress/-/issues/139).

Before
PHP warnings from plugins such as:

```
PHP Warning:  Attempt to read property "ID" on null in /path/to/httpdocs/wp-content/plugins/acf-extended/includes/module-post.php on line 39
```

After
No PHP warnings.

### DIFF
--- a/includes/admin-pages/civicrm.page.error.php
+++ b/includes/admin-pages/civicrm.page.error.php
@@ -66,7 +66,7 @@ class CiviCRM_For_WordPress_Admin_Page_Error {
     $this->add_menu_items($logo, $position);
 
     // Add our meta boxes.
-    add_action('add_meta_boxes', [$this, 'meta_boxes_error_add']);
+    add_action('civicrm/page/error/add_meta_boxes', [$this, 'meta_boxes_error_add']);
 
   }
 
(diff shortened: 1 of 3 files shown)


kcristiano:
@christianwach It read like a good change to me, but all PRs are failing test `E2E_Core_PathUrlTest/testGetUrl_WpAdmin/` here .  I am thinking we may need to ignore that one.

@seamuslee001 @totten any input?  I tried to fix the test but it passes sporadically on my buildkit local install

christianwach:
Is it [this test](https://github.com/civicrm/civicrm-core/blob/master/tests/phpunit/E2E/Core/PathUrlTest.php#L91-L117) that's failing? I don't have `phpunit` here, but looking at those `assertEquals()` statements, they shouldn't be failing given the output of `cv('url civicrm/contribute?reset=1')` on my machine. Weird.

christianwach:
@kcristiano Thanks for reviewing 👍

kcristiano:
test this please

PR #298, opened 2023-06-22
Fix UTC offsets when retrieving timezone string in WordPress

christianwach:
Overview
Fixes [this issue on Lab](https://lab.civicrm.org/dev/wordpress/-/issues/121).

Before
In WordPress, one can choose e.g. `UTC+4` as the site's timezone. This results in no value being returned by `get_option('timezone_string')` which may lead to unexpected results in CiviCRM.

After
When choosing e.g. `UTC+4` in WordPress, a deprecated (but still currently valid) `Etc/UTC-4` timezone string is used.

Comments
There is a companion PR in [the CiviCRM-Core repo](https://github.com/civicrm/civicrm-core/pull/26604).

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -1152,10 +1152,10 @@ class CiviCRM_For_WordPress {
      * WordPress has it's own timezone calculations. CiviCRM relies on the PHP
      * default timezone which WordPress overrides with UTC in wp-settings.php
      */
-    $wpBaseTimezone = date_default_timezone_get();
-    $wpUserTimezone = get_option('timezone_string');
-    if ($wpUserTimezone) {
-      date_default_timezone_set($wpUserTimezone);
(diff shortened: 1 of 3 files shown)


christianwach:
Gah, the old `Undefined array key "page"` test failure 🤷

demeritcowboy:
I did a quick little run. For example creating an activity in the UI defaults the activity date to the wrong time currently, and with the PR sets it according to the configured offset.

demeritcowboy:
jenkins retest this please

kcristiano:
Failing Check is unrelated.

Have done multiple `r-run` via API and different forms we use on client sites.  All looks good.

Will merge once https://github.com/civicrm/civicrm-core/pull/26604   is merged.

PR #300, opened 2023-08-12
Fix warning caused by passing null to strpos

colemanw:
Overview
Fixes a bunch of PHP warnings when you visit the CiviCRM home page (dashboard) without a `q=` argument in the URL.

Before
Home page looks like this:
image

After
Home page looks like this:
image

Technical Details
----------
There was no reason to return NULL in a variable that was supposed to be a string.

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -1423,11 +1423,11 @@ class CiviCRM_For_WordPress {
    *
    * @since 4.6
    *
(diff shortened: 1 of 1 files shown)


kcristiano:
This looks good and  `r-run` removes the error on php 8.1.  Also did an `r-run` on php 7.4 with no issues.

PR #301, opened 2023-08-23
Compatibility framework including URL support for Polylang

christianwach:
Overview
Solves [this issue on Lab](https://lab.civicrm.org/dev/wordpress/-/issues/143). Hopefully provides a better framework for solving #289 as well.

Before
When Polylang is active, locale is lost on "secondary" or "internal" URLs built by CiviCRM Core. For example, on an Event Info screen (`https://example.org/fr/civicrm/event/info/?reset=1&id=3`) the "Register" link does not retain the current front-end locale but instead contains no locale information at all (`https://example.org/civicrm/event/register/?id=3&reset=1`). 

After
The link becomes `https://example.org/fr/civicrm/event/register/?id=3&reset=1` and retains the locale as defined by Polylang.

Technical Details
Requires a [corresponding PR on CiviCRM-Core](https://github.com/civicrm/civicrm-core/pull/27128).

### DIFF
--- a/includes/civicrm.basepage.php
+++ b/includes/civicrm.basepage.php
@@ -37,6 +37,30 @@ class CiviCRM_For_WordPress_Basepage {
   public $civi;
 
   /**
+   * @var bool
+   * Base Page parsed flag.
+   * @since 4.6
+   * @access public
+   */
+  public $basepage_parsed = FALSE;
+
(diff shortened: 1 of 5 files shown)


colemanw:
Merging based on review/testing by @mlutfy

PR #303, opened 2023-09-08
Merge 5.66 into master

mlutfy:
Forward merge #301 into master.

cc @kcristiano @christianwach

### DIFF
--- a/includes/civicrm.basepage.php
+++ b/includes/civicrm.basepage.php
@@ -37,6 +37,30 @@ class CiviCRM_For_WordPress_Basepage {
   public $civi;
 
   /**
+   * @var bool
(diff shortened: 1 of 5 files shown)


kcristiano:
The Core PR that is related seems to be merged here - https://github.com/civicrm/civicrm-core/commit/a9fd7acb608c314f7f456a43ee1b121098e2cb67

Not sure if this is a valid test fail on `E2E\Core\PathUrlTest::testUrl`

test this please

colemanw:
I'm not sure what's going on in that test, I don't see anything that would have changed in that code to cause a failure, but if we don't merge this then wpmaster is broken, so merging.

PR #304, opened 2023-09-24
NFC - Fix breadcrumb variable type in docblock

colemanw:
Docblock says `string` but it is always an `array`.


kcristiano:
test this please

kcristiano:
test fails unrelated

PR #305, opened 2023-11-05
[REF][PHP8.2] Declare properties

braders:
Overview
Ensure properteies are declared, for PHP 8.2 compatiability.

Before
Dynamic properties are deprecated on PHP 8.2. CiviCRM-WordPress was using a handful of dynamic properties, causing deprecation notices.

After
Properties are declared.

Comments
I wasn't 100% sure whether to go `protected` or `public` for the `wp_get`, `wp_post`, `wp_cookie` and `wp_request` properties - happy to change them to public if that is deemed better.

I noticed most other properties in this repo have `@since` set in the docblocks, but I was not sure what was the appropriate value, so have not added the annotation to these new properties.

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -197,6 +197,34 @@ class CiviCRM_For_WordPress {
    */
   public $admin;
 
+  /**
+   * @var array
+   * Reference to the original $_GET value.
+   * @access protected
+   */
+  protected $wp_get;
+
+  /**
+   * @var array
+   * Reference to the original $_POST value.
+   * @access protected
+   */
+  protected $wp_post;
+
+  /**
+   * @var array
+   * Reference to the original $_COOKIE value.
+   * @access protected
+   */
+  protected $wp_cookie;
+
+  /**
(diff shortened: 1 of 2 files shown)


christianwach:
@braders Thanks for this - the same code is in a PR that I was about to open that applies a whole heap of codestyle updates. I decided on `private` for the properties since they shouldn't ever be used except to restore state after CiviCRM has been invoked. What's your opinion on this?

braders:
@christianwach I don't really have any opinion between `protected` and `private`. CiviCRM seems to use `protected` more frequently, which is why I used that.

I can see a use-case for people accessing `menu_page` if they wanted to re-arrange the WordPress menu - that's why I kept it `public`.  That said, there are other ways to reference the pages in a WordPress menu if you have to (`global $menu`), so I don't really mind it going private.

christianwach:
I'll merge this since it does what it says and we can argue over property declarations when i open the codestyle PR.

PR #306, opened 2023-11-10
[NFC] Apply sub-set of WordPress codestyle

christianwach:
Overview
At present, this repo enforces the `civilint` codestyle - which is great to keep us all sane. However, there are many best practices when it comes to code that interacts with WordPress that are not covered by `civilint` but which can be determined using WordPress rulesets for `phpcs`. This PR adds a suitably-configured `phpcs.xml` file and updates the main plugin files to comply with it.

Technical Details
To test this PR, install the [WordPress Coding Standards for PHP_CodeSniffer](https://github.com/WordPress/WordPress-Coding-Standards) and make sure that the rulesets are known to `phpcs`, e.g.:

```sh
% phpcs -i
The installed coding standards are MySource, PEAR, PSR1, PSR2, PSR12, Squiz, Zend, WordPress, WordPress-Core, WordPress-Docs and WordPress-Extra
```

You should see no errors or warnings when running `phpcs` from the root directory of the repo.

I realise this is quite a substantial PR, but most of the changes are simple and fall broadly into the following categories:

* help text for translators
* avoiding clashes with WordPress globals
* using appropriate WordPress functions where applicable
* strict equality/inequality checks
* avoiding `extract()`
* consistency in docblocks 

The changes to the code do not alter the functionality of CIviCRM in WordPress - they only apply a number of recommendations from the WordPress rulesets. You can comment out individual `<exclude-pattern>` declarations in `phpcs.xml` to see how compatibility with `civilint` has been implemented. Where rules are deliberately ignored or disabled in code, you will find inline `phpcs:ignore` or or `php:disable` declarations in the appropriate places.

### DIFF
--- a/includes/civicrm.admin.php
+++ b/includes/civicrm.admin.php
@@ -208,11 +208,12 @@ class CiviCRM_For_WordPress_Admin {
     }
 
     // Bail if we are on our installer page.
-    if ($screen->id == 'toplevel_page_civicrm-install') {
+    if ($screen->id === 'toplevel_page_civicrm-install') {
       return;
     }
 
     $message = sprintf(
+      /* translators: 1: Opening strong tag, 2: Closing strong tag, 3: Opening anchor tag, 4: Closing anchor tag. */
       __('%1$sCiviCRM is almost ready.%2$s You must %3$sconfigure CiviCRM%4$s for it to work.', 'civicrm'),
       '<strong>',
       '</strong>',
@@ -234,6 +235,7 @@ class CiviCRM_For_WordPress_Admin {
   public function run_installer() {
 
     // Set install type.
+    // phpcs:ignore WordPress.WP.CapitalPDangit.Misspelled
     $_GET['civicrm_install_type'] = 'wordpress';
 
     $civicrmCore = CIVICRM_PLUGIN_DIR . 'civicrm';
@@ -300,11 +302,12 @@ class CiviCRM_For_WordPress_Admin {
     }
 
     // Bail if we are on our error page.
-    if ($screen->id == 'toplevel_page_CiviCRM') {
+    if ($screen->id === 'toplevel_page_CiviCRM') {
       return;
     }
 
     $message = sprintf(
+      /* translators: 1: Opening strong tag, 2: Closing strong tag, 3: Opening anchor tag, 4: Closing anchor tag. */
       __('%1$sCiviCRM needs your attention.%2$s Please visit the %3$sInformation Page%4$s for details.', 'civicrm'),
       '<strong>',
       '</strong>',
@@ -407,7 +410,7 @@ class CiviCRM_For_WordPress_Admin {
      * including "civicrm.settings.php" will throw a fatal error if $civicrm_root
      * is wrong.
      */
-    if ($error == FALSE) {
+    if ($error === FALSE) {
       $this->error_flag = 'settings-include';
       $initialized = FALSE;
       return FALSE;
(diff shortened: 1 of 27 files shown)


kcristiano:
Thanks @christianwach I've done a quick read through and it looks good.

I'll do a more thorough one and test as well.

The test failure seems empty....   I'll look at that as well.

christianwach:
> I'll do a more thorough one and test as well.

@kcristiano Thanks, that would be great.

> The test failure seems empty.... I'll look at that as well.

*sigh* yeah something something jUnit something 🤷

totten:
The test-failure tldr is that (a) we recently switched to running JS tests on Chromium Headless (*instead of PhantomJS*) but (b) the test-job on this repo is getting assigned to an older test-runner (*where Chromium isn't working properly*). (It's a little more involved, but that's the gist.) There's some discussion on MM (`~infrastructure`). We'll be doing some updates to the test-jobs to address it.

totten:
civibot, test please

totten:
civibot, test this please

kcristiano:
civibot, test this please

kcristiano:
@totten Any suggestions on the test failures?  

I want this in early, so I am inclined to merge with the failure as I believe it's a false positive.

includes/civicrm.admin.php
```diff
@@ -558,6 +561,11 @@ public function settings_change($dao) {
    *
    * @since 5.28
    * @since 5.33 Moved to this class.
+   *
+   * @param string $html The auto-update markup.
+   * @param string $plugin_file The path to the plugin.
+   * @param array $plugin_data An array of plugin data.
+   * @return string $html The modified auto-update markup.
    */
   public function auto_update_prevent($html, $plugin_file, $plugin_data) {
```
altafsamnani, line 570:
```suggestion
  public function auto_update_prevent(string $html, string $plugin_file, array $plugin_data): string 
  {
```

includes/civicrm.shortcodes.php
```diff
@@ -650,7 +649,8 @@ public function get_context() {
    *
    * @since 4.6
    *
-   * @return string The overridden content.
+   * @param string $content The content.
+   * @return string $content The overridden content.
    */
   public function get_content($content) {
```
altafsamnani, line 655:
```suggestion
  public function get_content(string $content): string {
```

altafsamnani (changes requested):

includes/civicrm.admin.php
christianwach: @altafsamnani Thanks for the suggestion(s) - the plugin doesn't use type declarations anywhere as yet, nor does it enforce them via `declare(strict_types=1)`. To add them throughout the codebase seems like a great commit in its own right but IMO is outside the scope of this PR since type declarations aren't included in the `phpcs` codstyle checks. @kcristiano What do you think?

kcristiano: I agree

totten:
@kcristiano Yeah, I see the same `PathUrlTest` issue in the daily test matrix. It's pre-existing (not caused by this PR).

totten:
(Also, an aside - in the Github UI, I see 4x reports re: `CiviCRM-WordPress-{Test,Style,PR}` and `default`. The `default` is a leftover artifact from the prior test configuration. Unfortunately, I don't know a simple way to clean old test results with either Github UI or Jenkins UI. But it shouldn't appear on future PRs.)

PR #307, opened 2023-11-10
Fix component/shortcode not rendering

grandeljay:
Overview
The contributions component does not seem to render as a shortcode (`[civicrm component="contribution" id="1" mode="live"]`). It only works for me when I use the hijack option.

Technical Details
I'm not familiar with CiviCRM as a user and even less as a developer but this code is inside the `render_single` method. Can't we expect `$this->shortcode_markup[$post->ID]` to always be `1` and `$this->shortcode_in_post[$post->ID]` to subsequently be `0`? 
For some reason `$this->shortcode_in_post[$post->ID]` otherwise ends up being `2` and the shortcode content can not be found.

Comments
I'm sorry, this is rather a question than a fix (I think).

Why does `$this->shortcode_in_post[$post->ID]` evaluate as `2` for me even though there's just one shortcode on the entire page?

The code from this PR is working for me (right now).

### DIFF
--- a/includes/civicrm.shortcodes.php
+++ b/includes/civicrm.shortcodes.php
@@ -345,12 +345,7 @@ class CiviCRM_For_WordPress_Shortcodes {
         if (isset($this->shortcode_markup[$post->ID])) {
 
           // Set counter flag.
-          if (!isset($this->shortcode_in_post[$post->ID])) {
-            $this->shortcode_in_post[$post->ID] = 0;
-          }
-          else {
-            $this->shortcode_in_post[$post->ID]++;
-          }
+          $this->shortcode_in_post[$post->ID] = 0;
 
           // This Shortcode must have been rendered.
           return $this->shortcode_markup[$post->ID][$this->shortcode_in_post[$post->ID]];


christianwach:
> Why does `$this->shortcode_in_post[$post->ID]` evaluate as `2` for me even though there's just one shortcode on the entire page?

@grandeljay That does sound odd. Which theme are you using?

grandeljay:
I'm using "my own" theme. It's the underscores (`_s`) starter theme. I had not considered that.

Is the theme relevant to this?

christianwach:
@grandeljay Well it looks like something (either your theme or one or more plugins) is triggering calls to `do_shortcode()` multiple times.

What's going on in this plugin is that it needs to parse all `[civicrm]` shortcodes before a page, post or archive is rendered. It does this in the `prerender()` method where it populates the `shortcode_markup` array with the relevant content. The reason it does this is because CiviCRM can only be "invoked" once per request, so if there are multiple shortcodes it has to render a preview that does not invoke CiviCRM. When the shortcodes are encountered in the content, the prerendered content is returned if there are multiple shortcodes and if not, then CiviCRM is invoked.

I can't say why your setup triggers `do_shortcode()` more than once so you're going to have to log some backtraces to discover why this is happening for you.

FWIW, have you tried the `[civicrm]` shortcode in isolation - e.g. with no plugins active?

christianwach:
@grandeljay FYI here is a matrix of situations in which CiviCRM shortcodes have been tested: https://github.com/civicrm/civicrm-wordpress/pull/281#issuecomment-1282147993 - I guess you can see why I'm asking about your particular setup :)

grandeljay:
Thanks for the info! I'll check my theme and get back to you!

grandeljay:
I did some more digging and couldn't find anything in my theme. I did however find that `register_hooks` called called once (as it should, i presume)
https://github.com/civicrm/civicrm-wordpress/blob/271bf53de32bf9a809b23a69af9429d1608efdf9/includes/civicrm.shortcodes.php#L80

But `render_single` is called three times!

Any idea how this is possible?

grandeljay:
> FWIW, have you tried the [civicrm] shortcode in isolation - e.g. with no plugins active?

I should have started by trying this. I found the issue (see linked issue)!

grandeljay:
Neat! Thanks for everything!

PR #310, opened 2023-11-22
Add utility method to return a Base Page URL regardless of context

christianwach:
Overview
Adds a utility method to return a Base Page URL regardless of context. Fixes [this issue on Lab](https://lab.civicrm.org/dev/wordpress/-/issues/144).

Steps to reproduce

1. Create a function:

```php
function civicrm_test_foo() {
  if (!civi_wp()->initialize()) {
    return '';
  }
  $url = CRM_Utils_System::url("civicrm/event/foo", "reset=1", TRUE, NULL, TRUE);
  return $url;
}
```

2. Create a page template, e.g. `page-foo.php` and a WordPress Page that uses the template. 
3. Add a CiviCRM Shortcode to the Page content.
4. In the page template, add:

```php
<pre><?php echo civicrm_test_foo(); ?></pre>
```

You should see:

```
https://example.org/foo/?civiwp=CiviCRM&q=civicrm%2Fevent%2Ffoo&reset=1
```

This is because CiviCRM is set to "Shortcode Mode" by the presence of the Shortcode in the content.

After
1. Update the function:

```php
function civicrm_test_foo() {
  // You can use "civicrm_basepage_url()" if you prefer.
  $url = civi_wp()->basepage->url("civicrm/event/foo", "reset=1", TRUE, NULL, TRUE);
  return $url;
}
```

You should see:

```
https://example.org/civicrm/event/foo/?reset=1
```

If you have Polylang installed, you should see the link point to the appropriate Base Page for the language of the Page. When you're viewing a translated page (e.g. in German), then you should see:

```
https://example.org/de/civicrm/event/foo/?reset=1
```

Comments

This PR also cleans up another "URL getter" method `CiviCRM_For_WordPress::get_base_url()` because it's no longer necessary. Adds a deprecation notice for any external code that calls it.


kcristiano:
test this please

kcristiano:
@christianwach It looks like this PR needs a rebase )console log)  I tried to apply manually and it failed.   Can you take a look?

christianwach:
@kcristiano I have rebased but have no idea what the test failure means.

kcristiano:
@christianwach This does exactly what you've detailed.   

The failures are all in the E2E tests.   I am going to see if we can solve that  E2E error.   

@totten I'd like this merged before 5.69 RC is cut from master

christianwach:
@kcristiano Can you explain what the test failures mean? I cannot make head nor tail of the report.

The only functional change is [here](https://github.com/civicrm/civicrm-wordpress/pull/310/files#diff-658b6770b2c22c48065243ebf76f9cc2381aa22cacf9a7ddfa0957a4b25ea38eR504) and everything else merely offers new (as yet unused) functionality.

kcristiano:
@christianwach the test failuers are all on E2E test that involve URL patterns.  While the tests may be in error as this PR does make a change to the method to get the URL we may have to adjust the test or there could be a real issue.   

I suspect the former, but I'd rather delaying merging to be sure.  I do plan on targeting the RC once this is resolved.   I want this in 5.69.x

kcristiano:
I just re-ran the E2E test locally and have:

```
Tests: 1652, Assertions: 1611, Errors: 25, Failures: 10, Skipped: 619, Incomplete: 5.
```

kcristiano:
> @kcristiano Okay thanks, I'll have a look tomorrow - I have an idea what the cause might be.

@christianwach I don't think it's the code, but the tests.    Based on my `r-run` and review the code looks good.   I just don't want to bork the test results by merging now.

PR #313, opened 2023-12-08
Use base page URL function to build links in multiple Shortcodes

christianwach:
Overview
Follow up to #310. It occurred to me that now that the base page URL function is available, it can be used to generate the links to the Base Page when rendering multiple Shortcodes.

Before
Code that replicates what `CRM_Utils_System_WordPress::url()` does in a Base Page context generates the URL to the Base Page. Not Polylang-aware so all Base Page links point to the default locale.

After
Code replaced with call to base page URL function. As a bonus, this now works as expected with Polylang - i.e. Base Page URLs respect the locale of where the multiple Shortcodes are rendered.

### DIFF
--- a/includes/civicrm.shortcodes.php
+++ b/includes/civicrm.shortcodes.php
@@ -486,54 +486,13 @@ class CiviCRM_For_WordPress_Shortcodes {
     // Default to no class.
     $class = '';
 
-    // Access CiviCRM config object.
-    $config = CRM_Core_Config::singleton();
-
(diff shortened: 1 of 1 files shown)


kcristiano:
@christianwach do we need to target 5.69 with this PR?

kcristiano:
`r-run` is fine, needed improvement.

kcristiano:
jenkins test this please

kcristiano:
test failure unrelated

PR #316, opened 2024-02-19
Add WPML integration for proper CiviCRM language generation dev/wordpress#133

shaneonabike:
Overview
Fixes [dev/wordpress#133](https://lab.civicrm.org/dev/wordpress/-/issues/133)

Before

After

Translated pages are being shown. 🎆

Testing
- [x] ```Different domain per language```: Two domains WPML setup to redirect to say xyz-fr.org for French and xyz.org for English
- [x] ```Different languages in directories```: One domain with xyz.org/fr (French) and xyz.org (english)
- [x]  ```Different languages in directories```: One domain with xyz.org/fr (French) and xyz.org/en (english)
- [x] ```Different languages in directories```: One domain with xyz.org/fr (French) and xyz.org (english) but client changes the main civicrm page for each language like ```civicrm``` (English) and ```civicrm-fr``` for French
- [ ] ```Different languages added as a parameter```: (less likely used but :shrug: )  Example: xyz.org/?lang=fr

### DIFF
--- a/includes/compatibility/civicrm.wpml.php
+++ b/includes/compatibility/civicrm.wpml.php
@@ -106,15 +106,15 @@ class CiviCRM_For_WordPress_Compat_WPML {
 
   }
 
- /**
-  * Checks WPML for CiviCRM Base Page matches.
-  *
-  * @since 5.70
-  *
-  * @param bool $is_basepage TRUE if the Post ID matches the Base Page ID, FALSE otherwise.
-  * @param int $post_id The WordPress Post ID to check.
-  * @return bool $is_basepage TRUE if the Post ID matches the Base Page ID, FALSE otherwise.
-  */
+  /**
+   * Checks WPML for CiviCRM Base Page matches.
+   *
+   * @since 5.70
+   *
+   * @param bool $is_basepage TRUE if the Post ID matches the Base Page ID, FALSE otherwise.
+   * @param int $post_id The WordPress Post ID to check.
+   * @return bool $is_basepage TRUE if the Post ID matches the Base Page ID, FALSE otherwise.
+   */
   public function basepage_match($is_basepage, $post_id) {
 
     // Bail if this is already the Base Page.
@@ -136,18 +136,18 @@ class CiviCRM_For_WordPress_Compat_WPML {
 
   }
 
- /**
-  * Filters the CiviCRM Base URL for the current language reported by WPML
-  *
-  * Only filters URLs that point to the front-end/back-end, since WordPress admin URLs are
-  * rewritten by WPML.
-  *
-  * @since 5.70
-  *
-  * @param str $url The URL as built by CiviCRM.
-  * @param bool $admin_request True if building an admin URL, false otherwise.
-  * @return str $url The URL as modified by WPML.
-  */
+  /**
+   * Filters the CiviCRM Base URL for the current language reported by WPML
+   *
+   * Only filters URLs that point to the front-end/back-end, since WordPress admin URLs are
+   * rewritten by WPML.
+   *
+   * @since 5.70
+   *
+   * @param str $url The URL as built by CiviCRM.
+   * @param bool $admin_request True if building an admin URL, false otherwise.
+   * @return str $url The URL as modified by WPML.
+   */
   public function base_url_filter($url, $admin_request) {
 
     // Skip when not defined.
@@ -182,14 +182,14 @@ class CiviCRM_For_WordPress_Compat_WPML {
 
   }
 
- /**
-  * Setup the rewrite rules for WPML.
-  *
-  * @since 5.70
-  *
-  * @param bool $flush_rewrite_rules True if rules flushed, false otherwise.
-  * @param WP_Post $basepage The Base Page post object.
-  */
+  /**
+   * Setup the rewrite rules for WPML.
+   *
+   * @since 5.70
+   *
+   * @param bool $flush_rewrite_rules True if rules flushed, false otherwise.
+   * @param WP_Post $basepage The Base Page post object.
+   */
   public function rewrite_rules_wpml($flush_rewrite_rules, $basepage) {
 
     // Bail if WPML is not present.
@@ -257,14 +257,14 @@ class CiviCRM_For_WordPress_Compat_WPML {
     }
   }
 
- /**
-  * icl_ls_languages WPML Filter
-  * Rewrite all CiviCRM URLs to contain the proper language structure based on the WPML settings
-  *
-  * @since 5.70
-  *
-  * @param array $languages passed by WPML to modify current path
-  */
+  /**
+   * icl_ls_languages WPML Filter
+   * Rewrite all CiviCRM URLs to contain the proper language structure based on the WPML settings
+   *
+   * @since 5.70
+   *
+   * @param array $languages passed by WPML to modify current path
+   */
   public function rewrite_civicrm_urls_wpml($languages) {
 
     // Get the post slug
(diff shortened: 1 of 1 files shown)


shaneonabike:
## Testing

+ Install WPML plugiin
Selection_019

1. Translate the CiviCRM Page
Selection_020
6. Translate said page above
7. View live version of page

> Why WPML

> Long-term Support

includes/compatibility/civicrm.wpml.php
```diff
@@ -100,4 +106,264 @@ public function locale_filter($locale) {
 
   }
 
+ /**
```
mlutfy, line 109: I think you have 1 space here instead of 2, so the style tests are failing.

shaneonabike: Super funny because I just changed them to be shorter because it was complaining but I think I misunderstood the problem. Thanks for the guidance.

christianwach:
@mlutfy @shaneonabike I've just spent 1.5 hours this morning trying to get WPML and the code in this PR working and have failed as it seems that my WPML developer account has expired. I have emailed them, but I'm afraid I'm not going to be able to continue to review this on a voluntary basis as it seems to be a time black hole. Sorry - hopefully someone else is prepared to put the hours in as a volunteer.

includes/compatibility/civicrm.wpml.php
```diff
@@ -100,4 +106,264 @@ public function locale_filter($locale) {
...
+        foreach ($rewrite as $path) {
+          add_rewrite_rule(
+            '^' . $path . '([^?]*)?',
+            'index.php?page_id=' . $post_id . '&civiwp=CiviCRM&q=civicrm%2F$matches[1]',
+            'top'
+          );
+        }
+      }
+    }
+    $this->rewrites = $rewrites;
```
christianwach, line 252: @shaneonabike You're going to need to declare the dynamic properties otherwise PHP 8.2+ will write lots of warnings to the logs.

christianwach (changes requested):
Please see the Polylang class for code you can copy.

includes/compatibility/civicrm.wpml.php
```diff
@@ -100,4 +106,264 @@ public function locale_filter($locale) {
...
+   * This code was taken from the WPML source, because it couldn't be referenced directly
+   *
+   * @since 5.70
+   *
+   * @param object $sitepress reference to WPML class
+   */
+  private function wpml_get_language_param_for_convert_url($sitepress) {
+    if (!empty($sitepress)) {
+      if (isset($_GET['lang'])) {
+        return filter_var($_GET['lang'], FILTER_SANITIZE_STRING);
```
christianwach, line 355: `FILTER_SANITIZE_STRING` is [deprecated in PHP 8.1.0](https://www.php.net/manual/en/filter.filters.sanitize.php). This will write warnings to the logs. [These filters](https://wpml.org/faq/how-to-get-current-language-with-wpml/) might work instead.

christianwach (changes requested):

includes/compatibility/civicrm.wpml.php
shaneonabike: Oh boy! I thought I had declared that good catch thanks!

includes/compatibility/civicrm.wpml.php
shaneonabike: Thanks for that! I actually found a new [filter](https://wpml.org/faq/how-to-get-current-language-with-wpml/) to grab the current language which is returning the same result. Much more forward thinking then re-using their code which was annoying. I changed the code and it is still working for that.

kcristiano:
@shaneonabike have you had a chance to test this PR:  https://github.com/shaneonabike/civicrm-wordpress/pull/1

The review and code in this PR looks to be an alternative method to achieve the goal of this PR.  If you agree that Christian changes work for you, we can move to get it merged.

shaneonabike:
Hi @kcristiano 

Yes I was happy to see that @christianwach did some refactoring of my code. I wrote them to let them know that I'll probably test mid-week and then we can get this in. I have a few scenarios I want to try out just to make sure it's all functioning. I'll ping you when it's all said and done!

shaneonabike:
Hi @christianwach 

So far so good. Although I noticed something a few things while testing your latest changes against mine.

+ If I enter the CiviCRM admin and the language is set to _English_ then the links for ```Online Contribution (Test-drive)``` point to a French path rather than the path of the language of my current user. This is true for setting the languages via a separate domain
+ Basically, when I migrated from using languages in the URL (for all languages) back to the normal format of hiding the default language I could no longer view the default language civicrm contribution page. It works for the French version. It's pretty edge case but I think someone might change their settings at some point so probably something we should look at. The path moves from ```/civicrm/contribute/transact/?reset=1&action=preview&id=1``` to ```/civicrm/?reset=1&action=preview&id=1```.
+ Contribution Page listing via FormBuilder links for live/preview go to a wp-admin path but I don't think that's related to this patch and could be fixed later

It's strange with the second issue I can't seem to repair the URLs now. I see if I can debug what is happening further, but when I switched the default language to French it started working again. I outlined all the test scenarios at the top, and have a few left to do.

I don't mind merging this but I think we should figure out what is happening also.

shaneonabike:
@christianwach 

Also! I noticed that you removed the check I had to see if WPML was installed. Can I ask why you did that? I would presume that we wouldn't want to fire/register any of this code if the plugin isn't installed right? Or is this check already there farther up in the stack

shaneonabike:
Further debugging seems that the problem lies in ```basepage_canonical_url```. For some reason the ```$argdata = $this->civi->get_request_args();``` is blank when switching the language. When I load the French page it loads properly with the proper values here.

To be honest though, this is a pretty edge case because on the other domain I was testing it is working. It just happened when I added the language parameter and then removed it.

shaneonabike:
@christianwach Ok last review. Changing a French CiviCRM page path to ```civicrm-fr``` and then trying to use the Contribution page doesn't seem to work. I think you indicated that your patch resolves that or was that incorrect? To be honest, most people don't change that path in my experience and use the default WPML editor to translate things, but if your refactoring of my code was to handle that it's not working for me presently.

christianwach:
> Or is this check already there farther up in the stack

@shaneonabike Yes: https://github.com/shaneonabike/civicrm-wordpress/pull/1/files#diff-5839a80c9f1ccde53fe698fb7b6520082e2b0ee0e1a319ef6f2f9f9c039ca210R79-R82

Regarding your other comments, I'm struggling to understand the context of each test. Can you edit them to describe your WordPress "General Settings", WPML language settings and User Profile "Site Language" setting in each test? 

Also, can you make sure your Permalinks are refreshed after each change? WPML requires that when changes are made.

christianwach:
@shaneonabike Okay, here are some thoughts in the meantime...

> If I enter the CiviCRM admin and the language is set to English then the links for Online Contribution (Test-drive) point to a French path rather than the path of the language of my current user.

Are you sure you refreshed Permalinks? Also, the language as chosen in the WPML language switcher in the admin bar should determine the language of the links - not your User Profile language, which determines the language in which CiviCRM is displayed. I found renaming the site on a per language basis helped disambiguate this.

> This is true for setting the languages via a separate domain

Have you verified that WPML is correctly formatting the Permalink for the Base Page to point to the appropriate sub-domain? I didn't test in those circumstances because I assumed that WPML would do the right thing and it's a much more complicated setup.

> when I migrated from using languages in the URL (for all languages) back to the normal format of hiding the default language I could no longer view the default language civicrm contribution page

I can't reproduce this.

shaneonabike:
> > Or is this check already there farther up in the stack
> 
> @shaneonabike Yes: https://github.com/shaneonabike/civicrm-wordpress/pull/1/files#diff-5839a80c9f1ccde53fe698fb7b6520082e2b0ee0e1a319ef6f2f9f9c039ca210R79-R82
> 
> Regarding your other comments, I'm struggling to understand the context of each test. Can you edit them to describe your WordPress "General Settings", WPML language settings and User Profile "Site Language" setting in each test?

Fair request. I'll have to do that in a week because I'm away presently for a week on reading break vacation (sorry for the delay). I'll detail everything in the test scenarios above and redo the tests just to make sure I didn't do something weird. Also, I believe my language was already defaulted to English (since I test in that) for my profile, but I can check that too b/c it's a relevant point.
> 
> Also, can you make sure your Permalinks are refreshed after each change? WPML requires that when changes are made.

Ahh yes a good pointer as well. I'm pretty particular about saving permalinks, WPML warns you to resave, and I probably did it like 2-4 times when I started having issues. It's bit me in the past so it's my first go to to reset things.

shaneonabike:
Hi everyone (@kcristiano @christianwach ) !

I hope that this finds you well.

I merged the changes that @christianwach made to my repo, and updated the code to be in sync. I don't think there are any factors. I still don't fully understand the issue @christianwach was pointing out about language parameters in the actual URL (or whether that works), but to be honest I'm happy if we don't support this. We just need to document it somewhere I think.

Sorry for all the delays hopefully we can get this into the next release which will help out users using WPML.

Have a good one.

christianwach (approved):

kcristiano:
merging based on @christianwach review

shaneonabike:
Thanks for all your hard work against this guys I think it's a great contribution!

PR #317, opened 2024-03-23
Skip Base Page test when there is no Post object

christianwach:
Overview
My logs seem to have a lot of entries of the following kind:

```
PHP Warning:  Attempt to read property "ID" on null in /path/to/civicrm/includes/compatibility/civicrm.polylang.php on line 191
```

This PR introduces a check to avoid such situations.

Before
PHP Warnings.

After
No PHP Warnings.

### DIFF
--- a/includes/compatibility/civicrm.polylang.php
+++ b/includes/compatibility/civicrm.polylang.php
@@ -184,8 +184,13 @@ class CiviCRM_For_WordPress_Compat_Polylang {
       return $redirect_url;
     }
 
-    // Bail if this is not a Base Page.
+    // Bail if there is no Post object.
     $post = get_post();
+    if (!($post instanceof WP_Post)) {
+      return $redirect_url;
+    }
+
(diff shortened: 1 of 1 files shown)


christianwach:
Hmm, something wrong with the tests again? I can't see why this relatively trivial change would fail.

kcristiano:
```
E2E\Core\LocalizedDataTest::testLocalizedData
The German SQL should match the German pattern.
Failed asserting that false is true.

/home/homer/buildkit/build/build-3/web/wp-content/plugins/civicrm/civicrm/tests/phpunit/E2E/Core/LocalizedDataTest.php:39
/home/homer/buildkit/extern/phpunit9/phpunit9.phar:2307`

```

That's odd

kcristiano:
jenkins test this please

kcristiano:
Code is valid, will solve issues with 404 and polylang pages that are 'virtual/dummy' pages.

PR #319, opened 2024-04-06
Update readme.txt with tested up to and stable version

kcristiano:
Update readme.txt with tested up to and stable version

NFC - this is needed, but also looking to flag up test failures


christianwach:
No functional changes, so tests must be faulty. Merging.

PR #320, opened 2024-04-06
Ensure permissions are always checked

christianwach:
Overview
Moves the permissions check so it always runs. Will catch an unlikely edge case where `have_posts()` returns `0`. Better safe than sorry eh?


kcristiano:
Thanks @christianwach Good improvement.

Tested, works as it should.

Merging

PR #318, opened 2024-04-06
Check permission before invoking CiviCRM

christianwach:
Overview
----------------------------------------
Fixes [this issue](https://lab.civicrm.org/dev/core/-/issues/5127) on Lab.

Before
----------------------------------------
`<form action="https://example.org/wp-admin/admin.php?page=CiviCRM&q=civicrm%2Fevent%2Fregister" ... >`

After
----------------------------------------
`<form action="https://example.org/civicrm/event/register/" ... >`

Technical Details
----------------------------------------
The new URL-building code in Core ultimately queries `CRM_Core_Config::singleton()->userFrameworkFrontend` to decide what `current:` should resolve to. This is (weirdly) set when checking access permissions. Previous URL-building code did not need to check this config property, so the change causes Base Page form submissions to point to the CiviCRM backend.

Affects CiviCRM 5.72.0.

PR #321, opened 2024-04-07
573 perms

kcristiano:
https://github.com/civicrm/civicrm-wordpress/pull/318

Overview
Fixes [this issue](https://lab.civicrm.org/dev/core/-/issues/5127) on Lab.

Before
`<form action="https://example.org/wp-admin/admin.php?page=CiviCRM&q=civicrm%2Fevent%2Fregister" ... >`

After
`<form action="https://example.org/civicrm/event/register/" ... >`

Technical Details
The new URL-building code in Core ultimately queries `CRM_Core_Config::singleton()->userFrameworkFrontend` to decide what `current:` should resolve to. This is (weirdly) set when checking access permissions. Previous URL-building code did not need to check this config property, so the change causes Base Page form submissions to point to the CiviCRM backend.

Affects CiviCRM 5.72.0.

https://github.com/civicrm/civicrm-wordpress/pull/320

Moves the permissions check so it always runs. Will catch an unlikely edge case where have_posts() returns 0. Better safe than sorry eh?

https://github.com/civicrm/civicrm-wordpress/pull/319 

Update readme.txt with tested up to and stable version

NFC - this is needed, but also looking to flag up test failures

### DIFF
--- a/includes/civicrm.basepage.php
+++ b/includes/civicrm.basepage.php
@@ -406,6 +406,13 @@ class CiviCRM_For_WordPress_Basepage {
         // Determine if the current Post is the Base Page.
         $is_basepage = $this->is_match($post->ID);
 
+        // Check permission.
+        $denied = TRUE;
+        $argdata = $this->civi->get_request_args();
+        if ($this->civi->users->check_permission($argdata['args'])) {
+          $denied = FALSE;
+        }
+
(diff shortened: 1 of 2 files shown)


kcristiano:
PRs originally added to 5.72  Cherry picked to RC branch

PR #324, opened 2024-04-11
Fix Asset Builder links in Afform Shortcodes

christianwach:
Overview
Fixes [this issue](https://lab.civicrm.org/dev/core/-/issues/5131) on Lab.

Before
Asset Builder links rendered via Shortcode or Base Page and finally `invoke()`

After
Asset Builder links rendered as `non-page` assets via `invoke()` bypassing other logic.

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -1451,7 +1451,7 @@ class CiviCRM_For_WordPress {
      * unnecessary?
      */
     if (CRM_Utils_Array::value('HTTP_X_REQUESTED_WITH', $_SERVER) === 'XMLHttpRequest'
-        || ($argdata['args'][0] === 'civicrm' && in_array($argdata['args'][1], ['ajax', 'file']))
+        || ($argdata['args'][0] === 'civicrm' && in_array($argdata['args'][1], ['ajax', 'file', 'asset']))
         // phpcs:disable WordPress.Security.NonceVerification.Recommended
         || !empty($_REQUEST['snippet'])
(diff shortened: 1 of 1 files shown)


civicrm.php
```diff
@@ -1451,7 +1451,7 @@ public function is_page_request() {
      * unnecessary?
      */
     if (CRM_Utils_Array::value('HTTP_X_REQUESTED_WITH', $_SERVER) === 'XMLHttpRequest'
-        || ($argdata['args'][0] === 'civicrm' && in_array($argdata['args'][1], ['ajax', 'file']))
+        || ($argdata['args'][0] === 'civicrm' && in_array($argdata['args'][1], ['ajax', 'file', 'asset']))
```
colemanw, line 1454: Wow, what a beautifully awful hack.

Long-term, seems like we ought to include this sort of thing in our router metadata instead of *guessing* what kind of route it is based on string-matching.

christianwach: Goes waaaay back [to 4.4](https://github.com/civicrm/civicrm-wordpress/pull/42).

And yes, some other way would definitely be better 😆👍 

FWIW I had a thought on how to make it slightly less hacky in the short term... I'll open a PR where we can discuss further.

christianwach: @colemanw We seem to have had almost the [same conversation](https://github.com/civicrm/civicrm-wordpress/pull/40#issuecomment-59153582) 10 years ago 👻

PR #325, opened 2024-04-11
NFC: Rewrite "is_page_request" method for comment

christianwach:
Overview
Follow up to #324 to discuss how best to improve handling the distinction between CiviCRM "pages" and "non-pages".

I've refactored the `is_page_request()` method to make it a bit more obvious what's going on and added a filter that *could* be used to modify the outcome of the "non-page" checks if someone really wanted to do that.

I tend to agree with @colemanw that routes in CiviCRM should have metadata that determines this, but it may not even occur to Extension authors unless they test on WordPress and/or Joomla.

Anyway, at least having this PR here will keep eyes on the issue.

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -1419,30 +1419,27 @@ class CiviCRM_For_WordPress {
    *
    * @since 4.4
    *
-   * @return boolean True if request is for a CiviCRM page, false otherwise.
+   * @return boolean $is_page True if request is for a CiviCRM page, false otherwise.
    */
   public function is_page_request() {
 
     // Assume not a CiviCRM page.
-    $return = FALSE;
+    $is_page = FALSE;
 
-    // Kick out if not CiviCRM.
+    // Bail if no CiviCRM.
     if (!$this->initialize()) {
       return $return;
     }
 
-    // Get args.
+    // Get request args.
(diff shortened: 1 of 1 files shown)


civicrm.php
```diff
@@ -1419,52 +1419,76 @@ private function restore_wp_magic_quotes() {
...
-      // phpcs:disable WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+      // phpcs:ignore WordPress.Security.NonceVerification.Recommended, WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
       $html = isset($_GET['html']) ? wp_unslash($_GET['html']) : '';
-      // phpcs:enable WordPress.Security.NonceVerification.Recommended
-      // phpcs:enable WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
     }
 
     /*
      * FIXME: It's not sustainable to hardcode a whitelist of all of non-HTML
      * pages. Maybe the menu-XML should include some metadata to make this
```
colemanw, line 1447: Oh look, there's the same comment I made in the other PR, probably from our conversation about this 10 years ago 👴🏻

kcristiano:
@christianwach this looks good to me and I agree it's NFC.

merging.

PR #327, opened 2024-05-30
NFC: Fix since tags in WPML methods

christianwach:
Overview
The merged WPML compatibility code will go into CiviCRM 5.75, so the methods should note that.


kcristiano:
Thanks Christian

PR #329, opened 2024-08-21
Shortcodes outside content are never pre-rendered

christianwach:
Overview
The check to see if a Shortcode has already been rendered fails when the CiviCRM Shortcode is being rendered via `do_shortcode('[civicrm ...]')` _outside_ The Loop (e.g. in a template). These Shortcodes will never have been pre-rendered and cannot therefore be present in the Shortcode markup array.

Before
Shortcodes outside The Loop cause "Undefined array key" PHP Warnings to be logged, return early and fail to render.

To test this, add an Afform Shortcode to your theme's footer.php, e.g.

```php
do_shortcode( '[civicrm component="afform" name="afformQuickAddIndividual"]' );
```

Then make it render by putting the following in a plugin or `functions.php` of your theme:

```php
/**
 * Support CiviCRM Form Builder Shortcode when rendered with "do_shortcode()"
 */
function force_civi_scripts_to_load() {
  if ( civicrm_initialize() ) {
    \Civi::service( 'angularjs.loader' )->addModules( 'afformQuickAddIndividual' );
    add_action( 'wp_enqueue_scripts', array( civi_wp(), 'front_end_page_load' ), 100 );
  }
}
add_action( 'get_header', 'force_civi_scripts_to_load' );
```

Visit a page or post which has a CiviCRM Shortcode in it. The footer Shortcode fails  to render.

After
Shortcodes outside The Loop render as expected.

Comments
I have tested in a number of Shortcode scenarios, but would appreciate others doing further testing.

### DIFF
--- a/includes/civicrm.shortcodes.php
+++ b/includes/civicrm.shortcodes.php
@@ -345,6 +345,8 @@ class CiviCRM_For_WordPress_Shortcodes {
    */
   public function render_single($atts) {
 
+    global $post;
+
     // Do not parse Shortcodes in REST context for PUT, POST and DELETE methods.
     // Nonce is not necessary here.
     // phpcs:ignore WordPress.Security.NonceVerification.Missing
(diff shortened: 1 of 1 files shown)


kcristiano:
@totten @seamuslee001 PR is merge ready, it works as expected and should be merged.

Test fails look unrelated, but wanted to check in before I ignored the results.  Thoughts?

kcristiano:
jenkins test this please

kcristiano:
`r-run` is fine, code makes sense.

Merging as I'd like this in to the RC early.

Test fails look unrelated.

composerjk:
Finally upgraded from CiviCRM 5.77.1 to 5.78.4 and found that the shortcodes we were using stopped working. When I put back `includes/civicrm.shortcodes.php` from 5.77.1, the shortcodes worked again.

We use [Themeco's Cornerstone](https://theme.co/pro) pagebuilder with their Pro theme, using a shortcode like:
```
[civicrm component="contribution" id="1" action="transact" mode="live" hijack="0"]
```
in a Raw Content element.

Mentioned in [Mattermost ~wordpress](https://chat.civicrm.org/civicrm/pl/hem4geb19pb57fdd1kqzztpj5e).

christianwach:
@composerjk Try #337 which addresses this.

PR #331, opened 2024-10-08
Prevent CiviCRM resources from loading when not in CiviCRM admin

christianwach:
Overview
----------------------------------------
Fixes [this issue on Lab](https://lab.civicrm.org/dev/wordpress/-/issues/151).

Before
----------------------------------------
CSS and JS from CiviCRM is loading on non-CiviCRM pages across WordPress admin.

After
----------------------------------------
CSS and JS from CiviCRM is restricted to CiviCRM pages.

Comments
----------------------------------------
It looks like I may have accidentally left a global scope hook callback in the code in [this PR](https://github.com/civicrm/civicrm-wordpress/pull/228).

PR #334, opened 2024-10-10
revert 151-lab due to issues discovered with shortcode button

kcristiano:
Revert https://github.com/civicrm/civicrm-wordpress/pull/331

While the PR fixes the admin loading, any use of the civicrm shortcode modal (classic editor) will now fail.    Rather than create a new regression, reverting for now.

We will need to review the best way to load the assets (js and css) for WP as the current method will not easily cover the needs here.


kcristiano:
self-merging as this is a revert of #331

PR #335, opened 2024-10-21
Fix fatal error if source is unavailable

christianwach:
Overview
Simple syntax fix to prevent a fatal error in this rare/unlikely edge case.


christianwach:
@kcristiano Whilst upgrading CiviCRM via WP-CLI on a live site. I assume there was a request during the short time that the CiviCRM plugin directory was being replaced.

PR #336, opened 2024-11-12
(dev/core#5464) Add support for IFRAME URLs via `?_cvwpif=1`

totten:
Overview

Allow Civi-WP to act as the the provider of IFRAMEs. When embedding a Civi page as an IFRAME, it needs to opt-out of standard cookies, add an alternative session-management ("cosession"), and hide site-wide nav-items.

* Contributes to: https://lab.civicrm.org/dev/core/-/issues/5464
* Depends on: https://github.com/civicrm/civicrm-core/pull/31460 

Before

No option to generate links.

After

You can make an IFRAME referencing a Civi page with these steps::

1. Enable the `iframe` extension
2. Check the configuration (`civicrm/admin/setting/iframe`) to see which pages are allowed as IFRAMEs. (Add or remove items to taste.)
3. Creatiea regular link to the intended page
4. Append `?_cvwpif=1` to the link

For example, I used this for my own testing:

```html
<html>
<body>

<iframe src="http://wpmaster.127.0.0.1.nip.io:8001/civicrm/event/register?id=7&reset=1&_cvwpif=1" width="90%" height="500">

</body>
```

Technical Details

The URL formula in this PR is a bit different from the URL in Civi-Drupal/Civi-Backdrop. Compare `iframe.php` with `_cvwpif`:

* `http://dmaster.localhost` __`/iframe.php`__ `/civicrm/event/register?id=7&reset=1`
* `http://wpmaster.localhost/civicrm/event/register?id=7&reset=1` __`&_cvwpif=1`__

The key reason is that Drupal/Backdrop have centralized management of the session-cookie, and we can only influence its behavior by intervening at the top of the request. By contrast, WP seems to have sufficient hookability.

To work correctly, this depends on https://github.com/civicrm/civicrm-core/pull/31460. However, it should be safe to merge in either order.

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -257,6 +257,12 @@ class CiviCRM_For_WordPress {
         include_once CIVICRM_PLUGIN_DIR . 'wp-cli/wp-cli-civicrm.php';
       }
 
+      define('CIVICRM_IFRAME', self::$instance->is_iframe());
+      if (CIVICRM_IFRAME) {
+        // Must run before WP starts processing cookies.
+        self::$instance->activate_iframe();
+      }
+
       // Delay setup until 'plugins_loaded' to allow other plugins to load as well.
       add_action('plugins_loaded', [self::$instance, 'setup_instance']);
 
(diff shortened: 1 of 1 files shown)


JoeMurray:
Thanks, @totten !

@kcristiano I'm wondering if you would be a good person to review the session management approach here?

kcristiano:
@JoeMurray I can take a look - I am out of the office this week, but back next.

JoeMurray:
Thanks, @kcristiano .

civicrm.php
```diff
@@ -292,6 +298,42 @@ public function __wakeup() {
...
+
+  protected function activate_iframe(): void {
+    // By default, internal links should stay in the iframe.
+    $GLOBALS['civicrm_url_defaults'][]['scheme'] = 'iframe';
+
+    // Strict browsers (eg Safari) will quietly disregard cookies when loading a page within an IFRAME.
+    // But it's quite awkward to test behavior with two variables (browser-type and page-context).
+    // For consistent testing/UX, we force similar behavior for any request with IFRAME-style URL.
+
+    // Variant A: Disregard specific pieces
```
kcristiano, line 321: @totten Looking at this section I'd like to clarify:

- if vwe determine we are in an iframe https://github.com/civicrm/civicrm-wordpress/pull/336/files#diff-302f8b24886ed21286616e661003877a59e8f3ca6d6af0437574f0bdee5b0a10R261 which depends on https://github.com/civicrm/civicrm-wordpress/pull/336/files#diff-302f8b24886ed21286616e661003877a59e8f3ca6d6af0437574f0bdee5b0a10R309
- we then remove the three following 'cookies'

If I have that correct, how do we then reset them after form submission?  
-

totten: @kcristiano Good question. I think you point to the right bits of code (basically: the presence of `$_REQUEST['_cvwpif']` determines if the current request is part of an IFRAME).

In Variant A, it doesn't actually delete the cookies. From an HTTP POV, the cookies are still transmitted and valid. But within the IFRAME request, we skip the WordPress function which reads the cookies for authentication. So as far as WordPress is concerned, within that IFRAME request, the cookies are inert -- leaving the user as anonymous.

Then on any other request, the cookies can still behave normally.

(IIRC, for Civi-WP environments, variants B+C also preserve the cookie from HTTP POV.)

kcristiano: Thanks Tim.   As soon as I complete testing we can make the core PR merge ready.   

I've also asked @christianwach for their input.   

It'd be nice to get this merged just after you branch for the next RC

mattwire: Confirming tested and working on remote WordPress site (different server) from WordPress/CiviCRM site.

PR #337, opened 2024-11-13
Enable theme compatibility checks for Shortcode rendering

christianwach:
Overview
Fixes [this issue on Lab](https://lab.civicrm.org/dev/core/-/issues/5553)

Before
Rendering Shortcodes only checks `in_the_loop()` to see if the Shortcode has been pre-rendered. This fails for some themes.

After
There is a new setting which allows a choice of checking `in_the_loop()` or whether `the_content` is the filter that is currently running. One of these settings *should* work for the majority of themes. If neither works, then there is a `civicrm_theme_compatibility_mode` filter through which custom theme compatibility checks can be made.

Screenshot 2024-11-13 at 13 43 33

Technical Details
An example of when neither option works would be when trying to render a Shortcode in an ACF "WYSIWYG" Field. ACF mimics what WordPress does using the `acf_the_content` filter when rendering the "WYSIWYG" Field on the front end, so the following code *should* work:

```php
/**
 * Applies a custom theme compatibility check.
 *
 * @param bool $theme_check True if the check has passed, false otherwise.
 * @param bool $theme_mode The Shortcode Theme Compatibility setting.
 * @return bool $theme_check True if the custom check has passed, false otherwise.
 */
function my_civicrm_theme_compatibility_mode( $theme_mode, $theme_check ) {

  // Render if the ACF content filter is the current filter.
  $filter_check = ('acf_the_content' === current_filter()) ? TRUE : FALSE;
  if ($filter_check) {
    $theme_mode = TRUE;
  }

  return $theme_mode;

}
add_filter( 'civicrm_theme_compatibility_mode', 'my_civicrm_theme_compatibility_mode', 10, 2 );

```

Note that the above code would be dependent on Shortcode pre-rendering e.g. via the `civicrm_prerender_post_content` filter and CiviCRM resources having been properly added.

### DIFF
--- a/assets/js/civicrm.options.js
+++ b/assets/js/civicrm.options.js
@@ -166,15 +166,18 @@
       // Assign properties.
       me.basepage_submit = $('#civicrm_basepage_submit');
       me.shortcode_submit = $('#civicrm_shortcode_submit');
+      me.theme_submit = $('#civicrm_theme_submit');
       me.email_submit = $('#civicrm_email_submit');
       me.permissions_submit = $('#civicrm_permissions_submit');
       me.cache_submit = $('#civicrm_cache_submit');
       me.basepage_select = $('#page_id');
       me.shortcode_select = $('#shortcode_mode');
+      me.theme_select = $('#theme_compatibility_mode');
       me.email_select = $('#sync_email');
       me.permissions_select = $('#permissions_role');
       me.basepage_selected = me.basepage_select.val();
       me.shortcode_selected = me.shortcode_select.val();
+      me.theme_selected = me.theme_select.val();
       me.email_selected = me.email_select.val();
 
       // Set status of Base Page submit button.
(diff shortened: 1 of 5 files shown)


christianwach:
just a force-pushed civilint fix. Not sure why GitHub auto-closed the PR 🤷

mattwire:
Test fail looks unrelated:
```
E2E_CivicrmAdminUi_ManageGroupsTest::testManageGroups
Behat\Mink\Exception\ElementNotFoundException: Element matching css "tr[data-entity-id='5']" not found.
```

JadeGaunt:
Tested and can confirm this fixes the issue.

kcristiano:
jenkins test this please

kcristiano:
test this please

composerjk:
I also confirm that this patch works. 

**Using CiviCRM 5.78.4 on WordPress 6.6.2:** Tested first with the previous `current_action()` check, replacing it with the updated `current_filter()`. That worked. Then, applied the patch and verified, again. Worked, once I changed the **Shortcode Theme Compatibility** setting.

There should be a highlighted note, perhaps, in the release notes when this is released.

Should it be backported into a 5.79 release since this impacts 5.78 and 5.79? I'm currently running 5.78.4 with the single-line fix from before.

PR #341, opened 2025-02-06
Fix intermittent ExpiredToken error when Iframe patches applied

mattwire:
Overview
Since enabling iframe on WordPress the site has been getting occasional crashes on the admin backend with "Firebase\JWT\ExpiredException: Expired token".
Eg. I triggered it by clicking "save" on a settings page. Definitely not running in an iframe.

Before
Crashes sometimes

After
Shouldn't crash. Follows identical pattern to other entrypoints.

Technical Details
All other entrypoints ONLY define `CIVICRM_IFRAME` it is is TRUE/1 - eg. see https://github.com/civicrm/civicrm-core/blob/master/ext/iframe/Civi/Iframe/EntryPoint/Drupal8.php#L17

But the WordPress implementation always defines `CIVICRM_IFRAME` and sets it to either TRUE or FALSE. This is problematic because all consumers of `CIVICRM_IFRAME` check if it is defined and do not check the value - eg. see https://github.com/civicrm/civicrm-core/blob/master/ext/iframe/Civi/Iframe/Cosession.php#L50

```
Firebase\JWT\ExpiredException: Expired token"
  "code" => null
  "exception" => Civi\Crypto\Exception\CryptoException {#15031
    #message: "Firebase\JWT\ExpiredException: Expired token"
    #code: 0
    #file: "/var/www/vhosts/xxx.org/wp-content/plugins/civicrm/civicrm/Civi/Crypto/CryptoJwt.php"
    #line: 101
    #cause: null
    -_trace: null
    -errorData: array:1 [
      "error_code" => 0
    ]
    trace: {
      /var/www/vhosts/xxx.org/wp-content/plugins/civicrm/civicrm/Civi/Crypto/CryptoJwt.php:101 {
        Civi\Crypto\CryptoJwt->decode($token, $keyTag = 'SIGN')
        ›   // Keep our signature independent of the implementation.
        ›   throw new CryptoException(get_class($e) . ': ' . $e->getMessage());
        › }
      }
      /var/www/vhosts/xxx.org/wp-content/plugins/civicrm/civicrm/ext/iframe/Civi/Iframe/Cosession.php:204 { …}
      /var/www/vhosts/xxx.org/wp-content/plugins/civicrm/civicrm/ext/iframe/Civi/Iframe/Cosession.php:70 { …}
      /var/www/vhosts/xxx.org/wp-content/plugins/civicrm/civicrm/ext/iframe/Civi/Iframe/Cosession.php:147 { …}
```

Comments

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -257,8 +257,8 @@ class CiviCRM_For_WordPress {
         include_once CIVICRM_PLUGIN_DIR . 'wp-cli/wp-cli-civicrm.php';
       }
 
-      define('CIVICRM_IFRAME', self::$instance->is_iframe());
-      if (CIVICRM_IFRAME) {
+      if (self::$instance->is_iframe()) {
+        define('CIVICRM_IFRAME', 1);
         // Must run before WP starts processing cookies.
         self::$instance->activate_iframe();
       }
@@ -1270,7 +1270,7 @@ class CiviCRM_For_WordPress {
     }
 
     // Do the business.
-    if (CIVICRM_IFRAME && \Civi::service('iframe.router')->getLayout() !== 'cms') {
+    if (defined('CIVICRM_IFRAME') && CIVICRM_IFRAME && \Civi::service('iframe.router')->getLayout() !== 'cms') {
       \Civi::service('iframe.router')->invoke([
         'route' => implode('/', $argdata['args']),
         'printPage' => function ($content) {


kcristiano:
@mattwire This makes sense, but I have been unable to trigger an erro - Can you details the steps to reproduce?  I think I am stuck in the 'sometimes' part of the problem.

civicrm.php
```diff
@@ -257,8 +257,8 @@ public static function singleton() {
         include_once CIVICRM_PLUGIN_DIR . 'wp-cli/wp-cli-civicrm.php';
       }
 
-      define('CIVICRM_IFRAME', self::$instance->is_iframe());
-      if (CIVICRM_IFRAME) {
+      if (self::$instance->is_iframe()) {
+        define('CIVICRM_IFRAME', 1);
```
seamuslee001, line 261: @mattwire I think the test failures might relate to this not always being defined any more

totten:
I think the change makes sense. I would suggest approach inspired by Postel's Law:

1. (_Be strict in what we emit_) Yes, update this to strictly emit `CIVICRM_IFRAME` when true. 
2. (_Be forgiving in what we accept_) Also, update [here](https://github.com/civicrm/civicrm-wordpress/blob/1bc31a1897d8824fe8ba2b2c65e6f43bf57224b8/civicrm.php#L1273) to be like `defined('CIVICRM_IFRAME') && CIVICRM_IFRAME` (or if a helper function works, `CRM_Utils_Constant::value('CIVICRM_IFRAME')`)
3. (_Be forgiving in what we accept_) Also, (somebody/separate PR) should update [that lazy check](https://github.com/civicrm/civicrm-core/blob/master/ext/iframe/Civi/Iframe/Cosession.php#L50) to be use similar guard.

mattwire:
> @mattwire This makes sense, but I have been unable to trigger an erro - Can you details the steps to reproduce? I think I am stuck in the 'sometimes' part of the problem.

@kcristiano Unfortunately it seems fairly random (I assume that a token is actually expiring and because we've entered the iframe code because the value of the constant is not checked it assumes we should have a non-expired token). Assuming there is a timeout on the "token" that's probably why it triggers seemingly randomly.

civicrm.php
mattwire: Should be resolved now thankyou

mattwire:
> I think the change makes sense. I would suggest approach inspired by Postel's Law:
> 
> 1. (_Be strict in what we emit_) Yes, update this to strictly emit `CIVICRM_IFRAME` when true.

This PR

> 2. (_Be forgiving in what we accept_) Also, update [here](https://github.com/civicrm/civicrm-wordpress/blob/1bc31a1897d8824fe8ba2b2c65e6f43bf57224b8/civicrm.php#L1273) to be like `defined('CIVICRM_IFRAME') && CIVICRM_IFRAME` (or if a helper function works, `CRM_Utils_Constant::value('CIVICRM_IFRAME')`)

Now updated in this PR. Agree a helper might be nice but we could add that later.

> 3. (_Be forgiving in what we accept_) Also, (somebody/separate PR) should update [that lazy check](https://github.com/civicrm/civicrm-core/blob/master/ext/iframe/Civi/Iframe/Cosession.php#L50) to be use similar guard.

See https://github.com/civicrm/civicrm-core/pull/31997

kcristiano:
@mattwire I think we should do this against the RC - do you agree?

mattwire:
> @mattwire I think we should do this against the RC - do you agree?

Good idea. It only causes problems if you actually enable the iframe extension so not a huge issue for most people but worth targetting the RC as that's where the iframe code was added.

PR #344, opened 2025-05-25
Add new WP CiviCRM setting option that allows for controlling whether…

rbaugh:
## Automatically Sign In WP User setting

Depends on https://github.com/civicrm/civicrm-core/pull/32861

Overview
_This will add a new setting option in the WP CiviCRM settings screen which will allow the site to control whether or not the new WP user will be automatically logged in when their user account is created._

Before
_Currently, when a new WP user is created from a membership/contribution, after creating the WP User, it will automatically log the user in._

After
_With the new CiviCRM setting, 'Automatically Sign In User', you can now control whether you would like to allow the default behavior or prevent the auto login from happening._

Technical Details
_We have added new WP option that can ideally control whether or not a user will be automatically logged in after their user account is created. This new option requires a change with the CiviCRM Core Wordpress class that will leverage this new option._

### DIFF
--- a/assets/js/civicrm.options.js
+++ b/assets/js/civicrm.options.js
@@ -170,15 +170,18 @@
       me.email_submit = $('#civicrm_email_submit');
       me.permissions_submit = $('#civicrm_permissions_submit');
       me.cache_submit = $('#civicrm_cache_submit');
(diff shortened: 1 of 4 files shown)


kcristiano:
see https://github.com/civicrm/civicrm-core/pull/32861#issuecomment-2908519100

PR #345, opened 2025-05-29
NFC: Fix codestyle

christianwach:
Overview
Fixes a few codestyle and documentation issues with #344. 

Also fixes some (probably unused in practice) non-functional code.


kcristiano:
Thanks @christianwach  Looks good

PR #347, opened 2025-06-06
add composer.json

jackrabbithanna:
Overview
I'm working on a methodology to install CiviCRM into Wordpress with composer. This requires a composer.json in the repository.

@totten 

We've got a prototype system we are using internally, and wanting to work on this so anyone can do the same. 

We have a small prototype script that we run on composer install/update to create a symlink to civicrm-core in ./vendor/civicrm/civicrm-core/ in the Wordpress plugin directory.

We also needed to symlink ./vendor/civicrm/civicrm-core/packages to ./vendor/civicrm/civicrm-packages directory  

This may not be the best way to do this, we can continue to work on that. Suggestions?

```
<?php

$rootDirPath = dirname(dirname(__FILE__));
// assumes WP webroot is in ./web/
$customDir = $rootDirPath . '/web/wp-content/plugins/custom';
// we putting contrib plugins here
$contribDir = $rootDirPath . '/web/wp-content/plugins/contrib';

$civicrmDir = $contribDir . '/civicrm/civicrm';
$linkMappings = [
  $contribDir . '/civicrm/civicrm' => $rootDirPath . '/vendor/civicrm/civicrm-core',
  $rootDirPath . '/vendor/civicrm/civicrm-core/packages' => $rootDirPath . '/vendor/civicrm/civicrm-packages',
];
foreach ($linkMappings as $source => $target) {
  if (is_link($source)) {
    echo "Unlinking $source to $target." . PHP_EOL;
    unlink($source);
  }
  if (!is_link($source) && is_dir($target)) {
    symlink($target, $source);
    echo "Symlinked $source to $target." . PHP_EOL;
  }
  else {
    echo "Source directory not found: $target" . PHP_EOL;
  }
}
```
Save that script in client repo in ./scripts/symlink-civicrm.php

Then in the client composer.json invoke the script on install/update

```
    "scripts": {
        "post-install-cmd": [
            "php scripts/symlink-civicrm.php"
        ],
        "post-update-cmd": [
            "php scripts/symlink-civicrm.php"
        ]
    }

```

We also have in the "extra" section of composer.json
```
        "wordpress-install-dir": "web",
        "installer-paths": {
            "web/wp-content/plugins/mu-plugins/{$name}/": ["type:wordpress-muplugin"],
            "web/wp-content/plugins/contrib/{$name}/": ["type:wordpress-plugin"],
            "web/wp-content/themes/contrib/{$name}/": ["type:wordpress-theme"],
            "web/wp-content/uploads/civicrm/ext/{$name}/": ["type:civicrm-extension"]
        },
```

using https://packagist.org/packages/johnpbloch/wordpress for Wordpress Core for now

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,35 @@
+{
+  "name": "civicrm/civicrm-wordpress",
+  "type": "wordpress-plugin",
+  "description": "Open source constituent relationship management for non-profits, NGOs and advocacy organizations.",
+  "license": "AGPL-3.0",
+  "authors": [
+    {
+      "name": "Coleman Watts",
+      "role": "Product Manager"
+    },
+    {
+      "name": "Joshua Gowans",
+      "role": "Project Manager"
+    },
+    {
+      "name": "Mathieu Lutfy",
+      "role": "Infrastructure"
+    },
+    {
+      "name": "Tim Otten",
+      "role": "Software Architect"
+    },
+    {
+      "name": "CiviCRM Community",
+      "homepage": "https://civicrm.org"
+    }
+  ],
+  "require": {
+    "composer/installers": "^1 || ^2",
+    "civicrm/civicrm-core" : ">=5.21.0 || dev-master"
+  },
+  "extra": {
+    "installer-name": "civicrm"
+  }
+}


totten:
@jackrabbithanna OK, cool. 😄   So I guess there's a narrow topic (re: `civicrm-wordpress.git:composer.json`) and then a general topic (re: Civi-WP project-template and related bugs+fixes). 

Let's also CC @JSProffitt. I believe he's also been working on using Civi-WP-Composer (iirc, with the [Bedrock WordPress template](https://roots.io/bedrock/)).

## civicrm-wordpress.git:composer.json

This looks a lot like the one in `civicrm-drupal-8.git`, so I think there's a general presumption in its favor.

The main difference I see regards `civicrm-asset-plugin` (*hard requirement on D8-11 versus undeclared on WP*). I tend to agree more with this PR 👍.

i.e. There is no innate dependence from `civicrm-wordpress.git` (or `civicrm-drupal-8.git` for that matter) to `civicrm-asset-plugin`. But there is an... affinity? I guess the nature of that affinity differs:

* In Drupal-land, there's an official composer site-template, and there's settled-policy about how to split JS/CSS assets and PHP `vendor/`. _Practically speaking_, `c-a-p` is the best-supported solution for applying that to Civi. I interpret the `requires...c-a-p` as a _convenience_ to site-builders.
* In WP-land, there is no official composer site-template. AFAICS, there seems to be a fluctuating list of templates+adapters. Keeping this dependency loose improves the prognosis for coping with the fluctuation over the long-term. (Maybe one template works well with `c-a-p` today, while a second older template doesn't need anything like it, and a third future template needs something different.)

(This `composer.json` also inherits a dependence on `>= 5.21 || dev-master`. That dependency is bit peculiar, but I don't think it can be improved in a meaningful way. In any case... no one has complained about it to me, so it must be fine. 🙃)

I favor merging as-is, but we can leave a couple days in case others have comments.

## General, Civi with WP project templates

It's cool that you've posted a working POC!

Using the symlink reminds of something that Jeremy (@JSProffitt) was talking about -- where the installer didn't autodetect a path correctly. I'm not sure if they're the same issue. (Maybe he can post the draft patch which made it more automatic?)

> using https://packagist.org/packages/johnpbloch/wordpress for Wordpress Core for now

That sounds reasonable to me... but I don't really have any experience with the differences between `johnpbloch/wordpress`, `roots/wordpress`, `fancyguy/webroot-installer`, etc. so I can't say anything useful (like "Woohoo!" or "Ewww!"). I suspect you and Jeremy have probably looked at them more than me, so you might want to compare notes about their pro's/con's.

But maybe that should go in Gitlab issue? Regardless of which templates are better today, it seems like we still need this metadata for `civicrm-wordpress.git:composer.json`...

JSProffitt:
I agree with @jackrabbithanna. I think there's definitely improvement that could be made with the symlinking or ultimately removing the symlinking and I'd really like to see all of the secrets removed from civicrm.settings.php in favor of .env or secrets management on all platforms, but this is a good start.

jackrabbithanna:
@JSProffitt yeah the civicrm.settings.php the installer generates will have to be updated in some way.
Without the civicrm.settings.php in the POC template, the CiviCRM plugin installs, CiviCRM tables created, but then, bloop the civi pages don't load, complaining about not finding CiviCRM Core files. Its probably not too hard, just a couple more or different hardcoded paths. We can work on that!
My template is definitely not meant to be a final public product or anything official. Just a way to get started and us figuring it out.

jackrabbithanna:
Created https://lab.civicrm.org/dev/wordpress/-/issues/155 for more discussion

totten:
Sounds like a consensus on the start. Merged. Submitted to https://packagist.org/packages/civicrm/civicrm-wordpress.

PR #349, opened 2025-06-19
composer codebase, discover vendor directory for installation

jackrabbithanna:
Overview
Continued work making CiviCRM compatible with composer based installations
Reference: https://github.com/civicrm/civicrm-wordpress/pull/348

Before
CiviCRM cannot be installed with CiviCRM Core is downloaded with composer and located in the ./vendor directory

After
CiviCRM can be installed

### DIFF
--- a/includes/civicrm.admin.php
+++ b/includes/civicrm.admin.php
@@ -273,6 +273,35 @@ class CiviCRM_For_WordPress_Admin {
 
     }
 
+    $vendor_setup_paths = [];
+    $civicrm_plugin_dir_array = explode(DIRECTORY_SEPARATOR, CIVICRM_PLUGIN_DIR);
+    foreach ($civicrm_plugin_dir_array as $directory_component) {
+      $vendor_setup_paths[] = $directory_component;
+      $civicrm_core_path = implode(DIRECTORY_SEPARATOR, array_merge($vendor_setup_paths, ['vendor', 'civicrm', 'civicrm-core']));
+      $civicrm_setup_autoload_path = $civicrm_core_path . DIRECTORY_SEPARATOR . 'setup' . DIRECTORY_SEPARATOR . 'civicrm-setup-autoload.php';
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
This PR assumes that the [CiviCRM Asset Plugin](https://lab.civicrm.org/dev/civicrm-asset-plugin) is installed and has previously been used to copy css/js assets to the Wordpress CiviCRM plugin directory, in ./civicrm/core 

I've been experimenting with a Wordpress Bedrock template lately, see notes: https://lab.civicrm.org/dev/wordpress/-/issues/155#note_183616
, so for it, in the composer.json, "extra" section, I add
```
    "civicrm-asset": {
      "path": "web/app/plugins/civicrm/civicrm",
      "url": "/app/plugins/civicrm/civicrm" 
    },
```
Site developer can have the Wordpress plugin directory anywhere in the codebase they choose, and can update the above locations as needed.

includes/civicrm.admin.php
```diff
@@ -273,6 +273,35 @@ public function run_installer() {
 
     }
 
+    $vendor_setup_paths = [];
+    $civicrm_plugin_dir_array = explode(DIRECTORY_SEPARATOR, CIVICRM_PLUGIN_DIR);
+    foreach ($civicrm_plugin_dir_array as $directory_component) {
```
christianwach, line 278: @jackrabbithanna Would it be better to apply `array_reverse( $civicrm_plugin_dir_array )` here so that directory traversal doesn't start at the root directory of the server, but starts in the CiviCRM plugin directory and works upwards instead?

jackrabbithanna: You're right, I updated the algorithm to do just that.

kcristiano:
see comments on https://github.com/civicrm/civicrm-core/pull/32945 and https://github.com/civicrm/civicrm-wordpress/pull/348

This has been reviewed and tested and is good to merge.

kcristiano:
@jackrabbithanna If you can write up docs for this as we discussed.  I think adding to docs.civicrm.org would be good.

PR #353, opened 2025-09-30
WIP: Allow API v4 to be accessed via WP REST

christianwach:
Overview
Enables API v4 to be accessed via WordPress REST API.

Before
Unable to access API v4 via WordPress REST API.

After
API v4  can be accessed via WordPress REST API.

Technical Details
URLs need to be formatted properly, e.g.

```
https://example.org/wp-json/civicrm/v3/rest?entity=Contact&action=get&key=<site_key>&api_key=<api_key>&json={"select":["*"],"where":[["id","IN",[203,147]]],"limit":25,"version":4}
```

Note that `"version":4` is added to the JSON array in order to call API v4.

Comments
We may want to consider a new REST API namespace for these calls, e.g. `civicrm/v4`.

### DIFF
--- a/wp-rest/Controller/Rest.php
+++ b/wp-rest/Controller/Rest.php
@@ -122,7 +122,16 @@ class Rest extends Base {
     $params = apply_filters('civi_wp_rest/controller/rest/api_params', $this->get_formatted_api_params($request), $request);
 
     try {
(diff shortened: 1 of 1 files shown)


kcristiano:
@christianwach I think we can close as we have determined that this should not be needed.   Do you agree?

PR #357, opened 2026-02-11
(dev/core#6171) Form Builder - Suppress admin menu when using single-page access-tokens

totten:
Overview

In Form Builder (Afform), there is an option to create links using "Message Tokens" (aka "Secret links"). These include an access-token (JWT) for viewing a specific page as a specific person. This PR updates the rendering of that page on WordPress.

See: https://lab.civicrm.org/dev/core/-/issues/6171#note_197450

Use Case / Steps to Reproduce

* Main browser
    * Login as admin.
    * In "Form Builder", create a form (e.g. "My Profile"). Grant access to "Generic: Anyone with secret link". Allow "Placement" via "Message Token". Add some `Individual` fields (like first name and last name).
    * Pick a user/contact who has admin rights in WordPress.
    * Send an email to that person. Use this new mail token.
* Private browser
    * Open the link in a new window.
    * Look at the page.

Before

At the top of the page, it shows an navbar:

Screenshot 2026-02-10 at 7 39 46 PM

This shows things like "Customize", "Howdy, admin" (with user icon"), and chat/discussion counter. This implies that you are logged-in and can use those features. But not really -- you're only allowed to use the current page. Clicking anything in the menu will redirect to a login page.

After

At the top of the page, navbar is hidden.

Screenshot 2026-02-10 at 7 40 06 PM

Technical Details

In dev/core#6171, @kcristiano emphasized that CiviCRM should generally try to be polite about the presence of the nav-bar -- because other plugins might be customizing the usage of that menu. I tried to frame this narrowly -- so that in normal workflows (with real logins), it doesn't change anything.

We could go further and introduce a setting for the behavior. I started drafting an implementation here: https://github.com/civicrm/civicrm-wordpress/commit/9f61d17696e07f9d6841bf1434e00777bc5b05a5. I got a bit side-tracked with trying to display the setting on `page=civi_options`. If we actually do want the setting, then I can carry-on with that. (Or, it may be easier to expose the setting on `civicrm/admin/setting/uf`, but it'd need an extra PR.)

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -871,6 +871,8 @@ class CiviCRM_For_WordPress {
     // Register hooks for clean URLs.
     $this->register_hooks_clean_urls();
 
+    add_filter('show_admin_bar', [$this, 'suppress_menu_single_page']);
+
     if (!class_exists('CiviCRM_WP_REST\Plugin')) {
 
       // Set up REST API.
@@ -1645,6 +1647,43 @@ class CiviCRM_For_WordPress {
     else {
       return Civi::paths()->getUrl('[wp.frontend]/.', $absolute ? 'absolute' : 'relative');
     }
+  }
+
+  /**
+   * (dev/core#6171) Conditionally suppress admin navbar.
+   *
+   * Normally, other parties decide whether to show admin navbar.
+   *
+   * But there's an edge-case when:
+   * - Displaying a custom afform...
+   * - With a single-page access-token...
+   * - To a person who normally has permission to do admin-y things...
+   * - But isn't actually logged in
+   *
+   * Here, the navbar will tell the user that they're logged in, but they're
+   * not really logged-in, and the nav-links don't work in their normal way.
+   *
+   * @param $showAdminNav
+   *
+   * @return mixed|void
+   */
+  public function suppress_menu_single_page($showAdminNav) {
+    if (!$this->is_page_request()) {
+      return $showAdminNav;
+    }
+
+    if (!defined('_CIVICRM_FAKE_SESSION')) {
+      return $showAdminNav;
+    }
+
+    $scopes = explode(' ',
+      CRM_Core_Session::singleton()->get('authx')['jwt']['scope'] ?? ''
+    );
+    if (!in_array('afform', $scopes)) {
+      return $showAdminNav;
+    }
+
+    return FALSE;
   }
 
 }


kcristiano:
@GuillaumeSorel   @Guydn   Can you test this PR?

Thanks @totten I'll add this to one of our test sites and see hwo it interacts with the other plugins that manipulate the navbar.   Thanks for working on this.

GuillaumeSorel:
> @GuillaumeSorel @Guydn Can you test this PR?
> 
> Thanks @totten I'll add this to one of our test sites and see hwo it interacts with the other plugins that manipulate the navbar. Thanks for working on this.

ok Guy will provide a feed-back

totten:
> (@christianwach) Thanks for the distinction. Can you point me to where JWT auth is extended from CiviCRM to WordPress?

Good question. Here's an illustration of typical code-path: https://gist.github.com/totten/80f815fdc3b1b3345e42d946325c61ab

> I'm curious as to how `current_user_can()` is enabled (and why) because I assume that only `Permission::check(...)` is required.

Well, permissions+roles are generally integrated between CMS+Civi -- you can use tools/users/roles/permissions in either side to reference+modify the tools/users/roles/permissions on the other side. It's easier to keep track of that if they match-up. Hopefully, we can keep them matched-up. 🤞 

> FYI a lot of traditional themes present, for example, "Edit" buttons on content...

Yeah, there's a mixed-blessing in how the "page-token" feature (`PageTokenCredential`) is scoped. It's _semi-generic_:

* Page-tokens are only generated and accepted for Afform content.
* They aren't generated or accepted for WP content or Drupal content or Quickform content. 
* Hypothetically, if page-tokens worked with WP content, then the content "Edit" button would be weird.
* Hypothetically, page-tokens could be implemented for more rendering systems. But each needs work (esp re:AJAX+FORMs).
* But since it's not actually supported... maybe not a big problem?

totten:
@christianwach Since this feels new, you might want to look at some links about the general authentication framework ("authx"):

* Dev Guide: [Authentication](https://docs.civicrm.org/dev/en/latest/framework/authx)
* Web UI: "Administer > System Settings > Authentication"

> I'm curious as to how `current_user_can()` is enabled (and why) because I assume that only `Permission::check(...)` is required.

"Why?" can be a few different questions :wink: but perhaps this answer is the most clear-cut -- those two methods are the same thing. The general purpose+function of `CRM_Core_Permission::check()` is to delegate to the local user-framework (UF) and check its policies. `CRM_Core_Permission*` does have some extension-points and filters/formatters (e.g. rewriting `access foo` as `access_foo`), but those serve portability needs. At its core, its job is to delegate to the UF.

----

There's another angle (which isn't exactly your question, but I think it speaks to the gist)... when processing an authenticated request, we can set the active [CMS User and/or CiviCRM Contact](https://docs.civicrm.org/dev/en/latest/framework/authx/#identity). It's hard to make a universal statement of "always" or "never" set the CMS user -- it depends a lot on site-architecture/customizations/use-cases. Authx allows some tuning with the settings, e.g.

Screenshot from 2026-02-18 22-09-27

(That setting says: "If a request specifies HTTP parameter `?_authx=...`, then (never)/(sometimes)/(always) set the active `User`.")

In the "Authentication" settings page, you'll find several variations on this setting. But... right now... for Afform page-tokens specifically... it doesn't show a setting. That's _mostly_ happenstance. The general default is [`userMode=>optional`](https://github.com/civicrm/civicrm-core/blob/70eddf78de20dd05e7dce8d80b5222f44fe7af67/ext/authx/Civi/Authx/Authenticator.php#L208). Theoretically, the setting still exists (*just undeclared*), and you could say:

```
cv vset authx_afformpage_user=ignore
```

Is it a good idea to set that? :grimacing: Yesnomaybeornot? I think it depends on the design of the specific form in the secret-link. At minimum, which [security model](https://github.com/user-attachments/assets/91efec9d-cc27-4377-a44d-ef830f76b891) is used? (_Guess_: For "User-based" (RBAC), you really need the `User`. But for "Form-Based" (FBAC), you probably don't need the `User`.)

----

In any case, that seems like a separate thing from this patch. This patch (and that hypothetical patch) don't preclude or obsolete each other. (Even if FBAC forms can ignore the `User`... it would still be better for RBAC forms if we suppress the admin-menu when using single-page access-tokens.)

PR #360, opened 2026-05-01
Update Stable tag

kcristiano:
Overview
Update Stable tag to CiviCRM 6.14

Leave Tested to to 6.9.x

Comments
In testing both the RC (6.14) and master (6.15) against WordPress 7.0 we are seeing issues with checkbox fields due to the changes to the WP admin theme.   WP 7.0 is slated for Late May release (May 20th), so I wanted to have this release state that 6.14 is the Stable CiviCRM tag as well as be sure we have tested to 6.9.

Follow up will be an Issue detailing the places we see problems.

cc @christianwach

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -3,7 +3,7 @@ Contributors: needle
 Tags: civicrm, crm
 Requires at least: 6.3
 Tested up to: 6.9
-Stable tag: 6.12
+Stable tag: 6.14
 License: AGPL3
 License URI: http://www.gnu.org/licenses/agpl-3.0.html
 


seamuslee001:
Happy to merge this, I am also wondering about if we should be making any changes here @kcristiano https://github.com/civicrm/civicrm-core/blob/master/distmaker/dists/common.sh#L271 going forward

kcristiano:
@seamuslee001 we should.  I had that on my todo list and it fell off.

kcristiano:
closing as stale

PR #363, opened 2026-05-25
Enable Sample Data to be loaded when activating CiviCRM

christianwach:
Overview
@kcristiano Seems I forgot to include this in the last WP-CLI update.

Before
Cannot load Sample Data.

After
Able to load Sample Data.


kcristiano:
reviewed and tested.  Good to merge

PR #362, opened 2026-05-05
updates to support composer based installs and allow installing while codebase includes civicrm.settings.php template

jackrabbithanna:
Overview
----------------------------------------
This PR includes all the changes from my other contributions to provide composer support. 

Adds ability to include civicrm.settings.php in repo as a template and allowing install to work. This is necessary for PaaS environments with read only filesystems.

PR #365, opened 2026-05-31
do not require check for or creation of civicrm.settings.php to allow…

jackrabbithanna:
… using a template, do check if civicrm database tables exist in the civicrm.settings.php CIVICRM_DSN

Overview
Now that composer based installations are supported, a next step is leveling up support for automatic deployments to PaaS systems with read only file systems such as Upsun. 

For many years the CiviCRM Wordpress plugin installation has checked for whether a civicrm.settings.php exists to determine if CiviCRM was already installed, and if so, then stop the installation process. 

This PR checks to see if a CIVICRM_DSN is defined via a civicrm.settings.php file, and if so, checks to see if "civicrm_" tables exist in that database. If not, proceed with installation, if so, halt the installation. 

The PR only does this for composer based installations, that is when civicrm-core is discovered in a vendor directory. 
Standard installs are not affected.

Before
Installing CiviCRM on PaaS with civicrm.settings.php template provided in repo impossible.

After
Installation is possible

Technical Details
Interesting tangent: https://lab.civicrm.org/dev/wordpress/-/work_items/162
Will update this PR and/or template below based on that work.

Replaces: https://github.com/civicrm/civicrm-wordpress/pull/362

Comments
Full example composer template: https://github.com/Skvare/upsun-wordpress-bedrock-civicrm-template

Detailed blog article explaining this: https://github.com/jackrabbithanna/wordpress-bedrock-civicrm-upsun-blog/blob/main/civicrm-wordpress-bedrock-on-upsun.md

### DIFF
--- a/civicrm.php
+++ b/civicrm.php
@@ -103,6 +103,54 @@ else {
 }
 
 /**
+ * Determine whether CiviCRM's database schema has actually been bootstrapped.
+ *
+ * In Bedrock-style deployments, civicrm.settings.php may be shipped as a
+ * template (driven by env vars) before the database has ever been initialized.
+ * In that case CIVICRM_INSTALLED is TRUE but the schema is empty. Without this
+ * check, the activation hook would skip the installer redirect and admin
+ * bootstrap would later fail trying to read from a missing schema.
+ *
+ * Probes the civicrm_domain table on CIVICRM_DSN. Cached per-request.
+ */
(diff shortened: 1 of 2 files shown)


christianwach:
@jackrabbithanna This looks like it's getting quite complicated. Do you think it's worth considering a new class to encapsulate Composer compatibility?

